PR #2, opened 2020-11-19
Added minimal tests, doc strings and few tweaks

balbirthomas:
- Added two test cases
- Added stop event handler
- Added doc strings to charm class
- Ensured replica set reconfiguration failure is not ignored

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,6 +27,13 @@ PEER = "mongodb"
 
 
 class MongoDBCharm(CharmBase):
+    """A Juju Charm to deploy MongoDB on Kubernetes
+
+    This charm has the following features:
+    - Add one more MongoDB units
+    - Reconfigure replica set anytime number of MongoDB units changes
+    - Provides a database relation for any MongoDB client
+    """
     state = StoredState()
 
     def __init__(self, *args):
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -145,24 +175,44 @@ def on_update_status(self, event):
...
 
         if (
             self.unit.is_leader()
-            and self.state.mongodb_initialized
             and self.need_replica_set_reconfiguration()
         ):
-            self.mongo.reconfigure_replica_set(self.cluster_hosts)
+            try:
+                self.mongo.reconfigure_replica_set(self.cluster_hosts)
+            except Exception as e:
```
justinmclark, line 192: Can this be a more explicit exception? I think it's fine either way though.

src/charm.py
```diff
@@ -145,24 +175,44 @@ def on_update_status(self, event):
...
     ##############################################
 
+    # handles client relation for MongoDB
     def on_database_relation_changed(self, event):
+        """Connect to database client
+
+        Any MongoDB client is provided with the following information
+        - Is MongoDB in a replicated or unitary state
+        - Replica set URI
+        - Standalone URI
```
justinmclark, line 210: Is this true anymore? I thought there was no longer a standalone deployment.

justinmclark (approved):
Just a couple of questions. +1

src/charm.py
balbirthomas: You mean specify which exceptions to catch. If so I can try. But the intention was to trap any exception and for any exception type, to ensure that it is raised again so that the event can be deferred.

src/charm.py
balbirthomas: There is not standalone deployment configuation option, but the user can always choose to deploy one unit and effectively be in standalone mode. I did not remove the standalone URI to allow a client charm wants to use it to access a single unit of MongoDB.

src/charm.py
justinmclark: If there are multiple exceptions that could occur, then I don't think it's worth worrying about.

src/charm.py
justinmclark: Ah I see. I was assuming the standalone URI wouldn't work if mongo was a replica set.

Abuelodelanada:
Looks good to me although I still don't have all the knowledge about charms.

PR #4, opened 2021-02-11
No need to check if pod spec has changed before running set_spec

mthaddon:
There's no need to check if the pod spec has changed before running set_spec, as Juju will only apply it if things have changed.


balbirthomas (approved):
Thanks. I understand this means Juju does not set a new spec unless it has changed.

PR #5, opened 2021-03-02
Refactor charm structure

balbirthomas:
This commit refactors the MogoDB charm codebase. The purpose
of this refactor is to separate logically distinct components
into separate files, objects and modules, and to remove dead
unused code.


Abuelodelanada (approved):
Looks good to me

dstathis (approved):
Looks good. Only thing I'm wondering is so many things are properties when nothing else calls this code

balbirthomas:
> Looks good. Only thing I'm wondering is so many things are properties when nothing else calls this code

Thanks. Actually all properties are used, in code that is not part of this pull request. Actually this PR deletes code that is not used.

PR #6, opened 2021-03-03
Fix typo - accessing standalone uri

balbirthomas:
This commit fixes an error introduced by a recent refactor
that does not specify the right object on which to access
the standalone_uri property.

### DIFF
--- a/tests/test_charm.py
+++ b/tests/test_charm.py
@@ -45,6 +45,29 @@ class TestCharm(unittest.TestCase):
                  'mongodb-1.mongodb-endpoints']
         mock_reconf.assert_called_once_with(peers)
 
+    def test_uri_data_is_generated_correctly(self):
+        self.harness.set_leader(True)
+        standalone_uri = self.harness.charm.mongo.standalone_uri
+        replica_set_uri = self.harness.charm.mongo.replica_set_uri
+        self.assertEqual(standalone_uri, 'mongodb://mongodb:27017/')
+        self.assertEqual(replica_set_uri, 'mongodb://mongodb-0.mongodb-endpoints:27017/')
+
+    def test_database_relation_data_is_set_correctly(self):
+        self.harness.set_leader(True)
+        rel_id = self.harness.add_relation('database', 'client')
+        self.harness.add_relation_unit(rel_id, 'client/1')
+        rel = self.harness.framework.model.get_relation('database', rel_id)
+        unit = self.harness.framework.model.get_unit('client/1')
+        self.harness.charm.on['database'].relation_changed.emit(rel, unit)
+        got = self.harness.get_relation_data(rel_id, self.harness.framework.model.unit.name)
+        expected = {
+            'replicated': 'False',
+            'replica_set_name': 'rs0',
+            'standalone_uri': 'mongodb://mongodb:27017/',
+            'replica_set_uri': 'mongodb://mongodb-0.mongodb-endpoints:27017/'
+        }
+        self.assertDictEqual(got, expected)
+
 
 def replica_set_name(pod_spec):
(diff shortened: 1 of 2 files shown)


Abuelodelanada (approved):

tests/test_charm.py
```diff
@@ -45,6 +45,29 @@ def test_replica_set_is_reconfigured_when_peer_joins(self, mock_reconf):
...
+        standalone_uri = self.harness.charm.mongo.standalone_uri
+        replica_set_uri = self.harness.charm.mongo.replica_set_uri
+        self.assertEqual(standalone_uri, 'mongodb://mongodb:27017/')
+        self.assertEqual(replica_set_uri, 'mongodb://mongodb-0.mongodb-endpoints:27017/')
+
+    def test_database_relation_data_is_set_correctly(self):
+        self.harness.set_leader(True)
+        rel_id = self.harness.add_relation('database', 'client')
+        self.harness.add_relation_unit(rel_id, 'client/1')
+        rel = self.harness.framework.model.get_relation('database', rel_id)
```
jameinel, line 59: I'm curious why you chose to go this route vs
```
self.harness.update_relation_data(rel_id, 'client/1', {'foo': 'bar'})
```

(FWIW, I'm a little surprised this is only triggered by relation-changed, and not relation-joined or even relation-created if you do have the information available)

dstathis (approved):

jameinel (approved):

tests/test_charm.py
balbirthomas: I did not want to set the data myself. I wanted the `mongo` object of the charm to generated the data. I guess I could have called the mongo object directly (as I did in the other test), generated the data and then set it.

I am not sure I have understood the second part of your question. I only have a database relation changed handler, so I am specifically testing it.

jameinel: I think you're missing the data model slightly. "relation-db-changed" is telling you that "the unit that I am associated with has new data for me" and that gives you the opportunity to reply with "based on the new information you gave me, here is some new data for you".
Calling "update_relation_data(..., 'client/1', ...)" is saying that client/1 has new data to give to mongo, and thus mongo can respond.

I would actually expect (based on the data that you are setting), that you could set all of this information in 'relation_created'. (I know have a relation to 'client', but I don't care about data for client/0 vs client/1, I can just tell them the mongo_uri that they should use.)
So in response to "you have a relationship with a new application" (relation-created) you can tell them "here is the URI you can use to talk to my application" (relation-set).

Harness.update_relation_data is not meant for the test suite to set values on behalf of self.unit. It is meant to set values on behalf of the other units that are related to this unit.

balbirthomas: That was illuminating. Thank you @jameinel . For this specific test I wanted to check that the harness unit (self.unit) is providing mongodb connection data to "client/1". As you mention "harness.update_relation_data" would indeed not serve my purpose. 

You make a good point on why I do not set mongodb information on relation joined. Let me think that through. I presume you meant relation joined (since I see no relation created in ops documentation). Only reason I set it in relation changed was to delay setting that data, because mongodb is slow to start and the administrator may add relations even before mongodb is fully functional.

balbirthomas: That was illuminating. Thank you @jameinel . For this specific test I wanted to check that the harness unit (self.unit) is providing mongodb connection data to "client/1". As you mention "harness.update_relation_data" would indeed not serve my purpose. 

You make a good point on why I do not set mongodb information on relation joined. Let me think that through. I presume you meant relation joined (since I see no relation created in ops documentation). Only reason I set it in relation changed was to delay setting that data, because mongodb is slow to start and the administrator may add relations even before mongodb is fully functional.

PR #20, opened 2021-09-24
Switch to using can connect pebble API

balbirthomas:
This commit guards pebble access using the new can_connect()
method. It also traps path and protocol errors when using the
push and pull pebble APIs.

These changes are being committed becasue ops.pebble.ChangeError
has been sporadically observed when there was an attempt to
execute container.restart()


src/charm.py
```diff
@@ -75,14 +76,15 @@ def _on_config_changed(self, event):
...
         container = self.unit.get_container("mongodb")
 
-        # Set security key
-        try:
-            if not self.have_security_key(container):
-                self.set_security_key(container)
-        except ConnectionError:
+        if not container.can_connect():
             self.unit.status = WaitingStatus("Waiting for Pebble ready")
+            event.defer()
```
sed-i, line 81: Defer does not eliminate to possibility that the charm would have to wait for update_status.
Unless we now know for sure that during startup, config_changed is guaranteed to fire before other events (pebble_ready, start). IIRC, during early work on alertmanager I occasionally observed config_changed firing last.

src/charm.py
```diff
@@ -331,11 +333,14 @@ def set_security_key(self, container):
...
-                       user="mongodb",
-                       group="mongodb")
+        try:
+            container.push(file_path,
+                           self.security_key,
+                           permissions=0o400,
+                           user="mongodb",
+                           group="mongodb")
+        except (PathError, ProtocolError):
+            self.unit.status = BlockedStatus("Failed to set security key")
```
sed-i, line 343: Consider adding logger.error with file_path.

src/charm.py
balbirthomas: Agreed. I will remove the defer() and retest. The Problem I was seeing with Juju is no longer reproducible so I can not be sure if this defer would solve any problem.

src/charm.py
balbirthomas: logger.error is only obvious if the adminstrator chooses to look at the logs. More than likely the admin will look at juju status. For this reason I would prefer to leave it at BlockedStatus. I can add a logger.error additionally.

src/charm.py
balbirthomas: Removed defer and retested (using mongoconsumer).

Issue #16, opened 2021-09-08
Redeploying mongdb in the same model leads to authentication failure.


balbirthomas:
## Steps to reproduce

1. Deploy mongodb charm
2. Use `juju remove-application` to remove the deployed mongodb-charm
3. Deploy mongodb charm again

This leads to an authentication failure error

```
unit-mongodb-k8s-0: 17:30:08 ERROR unit.mongodb-k8s/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/pymongo/pool.py", line 1394, in _get_socket
    sock_info = self.sockets.popleft()
IndexError: pop from an empty deque

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./src/charm.py", line 347, in <module>
    main(MongoDBCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/ops/main.py", line 409, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/ops/main.py", line 143, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/ops/framework.py", line 278, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/ops/framework.py", line 722, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/ops/framework.py", line 767, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 105, in _on_config_changed
    self._on_update_status(event)
  File "./src/charm.py", line 160, in _on_update_status
    if not self.mongo.is_ready():
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/src/mongoserver.py", line 54, in is_ready
    client.server_info()
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/pymongo/mongo_client.py", line 1994, in server_info
    return self.admin.command("buildinfo",
  File "/var/lib/juju/agents/unit-mongodb-k8s-0/charm/venv/pymongo/database.py", line 751, in command
    with self.__client._socket_for_reads(
[...]

PR #37, opened 2022-02-04
Moved key file from persistent to container storage

balbirthomas:
The MongoDB charm was initaly built as a pod spec charm. The
mongod daemon expects its key file to be present before it is
started. Hence the pod spec charm used a Kubernetes init container
and a persistent storage to create and populate the key file.
This is no longer necessary for a side car charm, since key files
can be pushed to the workload container before starting the
workload. This is exactly the workflow implemented by this

Closes: #16

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -16,8 +16,6 @@ containers:
     mounts:
       - storage: db
         location: /data/db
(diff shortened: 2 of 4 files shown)


LICENSE
```diff
@@ -1,674 +1,202 @@
...
-For more information on this, and how to apply and follow the GNU GPL, see
-<http://www.gnu.org/licenses/>.
-
-  The GNU General Public License does not permit incorporating your program
-into proprietary programs.  If your program is a subroutine library, you
-may consider it more useful to permit linking proprietary applications with
-the library.  If this is what you want to do, use the GNU Lesser General
-Public License instead of this License.  But first, please read
-<http://www.gnu.org/philosophy/why-not-lgpl.html>.
+
```
simskij, line 1: Change not within the scope of the PR

simskij (changes requested):

LICENSE
balbirthomas: Ok. Removing license change and pushing now.

simskij (approved):

jnsgruk (approved):
LGTM, thanks @balbirthomas

PR #68, opened 2022-11-14
rework dependency management

delgod:


pyproject.toml
```diff
@@ -24,22 +24,28 @@ tenacity = "^8.0.1"
...
+pyproject-flake8 = "^5.0.4.post1"
+pep8-naming = "^0.13.2"
+codespell = "^2.2.2"
+
+[tool.poetry.group.unit.dependencies]
+coverage = {extras = ["toml"], version = "^6.5.0"}
+pytest = "^7.1.1"
+
+[tool.poetry.group.integration.dependencies]
+juju = "^2"
```
marcoppenheimer, line 45: Are we using semver 2 for a reason? Safety against changes from 3.0 bugs?

tests/unit/test_charm.py
```diff
@@ -507,11 +507,11 @@ def test_reconfigure_remove_member_failure(self, connection, defer):
...
 
             # simulate 2nd MongoDB unit joining( need a unit to join before removing a unit)
-            self.harness.add_relation_unit(rel.id, "mongodb/1")
-            self.harness.update_relation_data(rel.id, "mongodb/1", PEER_ADDR)
+            self.harness.add_relation_unit(rel.id, "mongodb-k8s/1")
+            self.harness.update_relation_data(rel.id, "mongodb-k8s/1", PEER_ADDR)
 
             # simulate removing 2nd MongoDB unit
-            self.harness.remove_relation_unit(rel.id, "mongodb/1")
+            self.harness.remove_relation_unit(rel.id, "mongodb-k8s/1")
```
marcoppenheimer, line 514: Not worth adding in this PR, but worth adding a global `CHARM_KEY` or similar to avoid this.

marcoppenheimer (approved):
Looks good, left a couple non-blocking points/Qs. 
I might add `*.lock` to the `.gitignore` to spare diffs, as I *think* it will build the lock on the first `tox` run.

pyproject.toml
delgod: Looks like charms built on libjuju 3 cannot work on juju 2 - I see this issue https://bugs.launchpad.net/juju/+bug/1995131
Unfortunately, we can't move to juju 3 because of scale down issue - https://bugs.launchpad.net/juju/+bug/1977582

so I hardcoded juju 2.x

delgod:
> Looks good, left a couple non-blocking points/Qs. I might add `*.lock` to the `.gitignore` to spare diffs, as I _think_ it will build the lock on the first `tox` run.

I intentionally commit lock files to ensure that we will have the same libraries on my machine during development, in github tests and later on the customer system.
Such an approach allows us to have (kind of) Reproducible Builds, please read https://reproducible-builds.org/.

tests/unit/test_charm.py
delgod: separate PR should be good, thanks

pyproject.toml
marcoppenheimer: Yup, thought so, thanks for confirming.

marcoppenheimer (approved):

marcoppenheimer:
> > Looks good, left a couple non-blocking points/Qs. I might add `*.lock` to the `.gitignore` to spare diffs, as I _think_ it will build the lock on the first `tox` run.
> 
> I intentionally commit lock files to ensure that we will have the same libraries on my machine during development, in github tests and later on the customer system. Such an approach allows us to have (kind of) Reproducible Builds, please read [reproducible-builds.org](https://reproducible-builds.org/).

Of course, I missed you were using caret versions rather than specific X.Y.Z. Makes sense.

deusebio (approved):
Looks good!

Yes, I would upvote for commiting lock file. I guess the main idea of blocking that is that we would have all the same environment. 

However this would make it somewhat redundant the "requirements.txt", which can always be generated from the lock file, but I suppose we cannot skip that as it is required by charmcraft. 

If we had some "build" shortcut (enabled via make or tox), we could treat requirements.txt as a "compiled" version

PR #71, opened 2022-11-22
HA tests - scale up capablities, scale down capabilities, replication across members, replication member scaling and restart db process tests and CI housekeeping

dragomirp:
# Issue
* Porting more of the HA VM tests to the K8S charm

# Solution
* Ported add units, scale down capabilities, replication across members, replication member scaling and restart DB process tests and necessary utility functions
* Updated the microk8s version and checkout version action for CI

# Context
* To verify that the db process restart test steps down, election logging verbosity is increased and hunt for the step down message in the mongod conitainer logs in K8S
* Changes the way HA tests access replica set status and config
* PyMongo clients in the tests should be used as context managers, to clean themselves up
* Moved getting the total writes into a separate utility function to reduce boilerplate
* Updating microk8s and checkout action is not necessary, but everything seems to be working fine with the new versions

# Testing
* add units integration test
* scale down capabilities integration test
* replication across members integration test
* replication member scaling integration test
* restart db process integration test

# Release Notes


.github/workflows/ci.yaml
```diff
@@ -35,13 +35,12 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup operator environment
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: microk8s
           # This is needed until https://bugs.launchpad.net/juju/+bug/1977582 is fixed
-          channel: 1.23/stable
```
dragomirp, line 44: Everything seems to work on 1.25

tests/integration/ha_tests/test_ha.py
```diff
@@ -87,11 +227,10 @@ async def test_kill_db_process(ops_test: OpsTest, continuous_writes):
...
     # wait
-    client = await get_mongo_client(ops_test, excluded=[primary])
-
-    writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
-    time.sleep(5)
-    more_writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
+    with await get_mongo_client(ops_test, excluded=[primary]) as client:
+        writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
+        time.sleep(5)
+        more_writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
```
dragomirp, line 233: To clean up the client

tests/integration/ha_tests/helpers.py
```diff
@@ -274,25 +279,18 @@ async def mongod_ready(ops_test: OpsTest, unit: int) -> bool:
 
 async def get_replica_set_primary(ops_test: OpsTest) -> str:
     """Returns the primary unit name based no the replica set host."""
-    rs_status = await run_mongo_op(ops_test, "rs.status()")
-    assert rs_status.succeeded, "mongod had no response for 'rs.status()'"
+    with await get_mongo_client(ops_test) as client:
+        data = client.admin.command("replSetGetStatus")
```
dragomirp, line 283: `run_mongo_op()` was very flakey on my machine. Is it OK to just call the units directly instead?

tests/integration/ha_tests/test_ha.py
```diff
@@ -50,6 +58,16 @@ async def continuous_writes(ops_test: OpsTest) -> None:
     await clear_writes_action.wait()
 
 
+@pytest_asyncio.fixture
+async def change_logging(ops_test: OpsTest):
+    """Increases and resets election logging verbosity."""
+    await set_log_level(ops_test, 5, "replication.election")
```
dragomirp, line 64: Couldn't get the log message locally without increasing the verbosity.

tests/integration/ha_tests/test_ha.py
```diff
@@ -65,6 +83,128 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+    expected_units = len(ops_test.model.applications[app].units) - minority_count
+
+    # find leader unit
+    leader_unit = await find_unit(ops_test, leader=True)
+
+    # verify that we have a leader
+    assert leader_unit is not None, "No unit is leader"
+
+    # Force delete the leader and scale down
+    await kubectl_delete(ops_test, leader_unit, False)
```
dragomirp, line 128: The VM test destroys the leader, but in K8S we can't control which unit to scale down. I can delete it manually, to mimic the behaviour, but the StatefulSet would most probably just revive it. Do we need this?

tests/integration/ha_tests/test_ha.py
```diff
@@ -65,6 +83,128 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+    with await get_mongo_client(ops_test, excluded=original_units) as client:
+        db = client["new-db"]
+        test_collection = db["test_ubuntu_collection"]
+        query = test_collection.find({}, {"release_name": 1})
+        assert query[0]["release_name"] == "Focal Fossa"
+
+        # verify that the no writes were skipped
+        total_expected_writes = await get_total_writes(ops_test)
+        actual_writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
+        assert total_expected_writes == actual_writes
```
dragomirp, line 205: Should the extra unit be scaled down once the test is done?

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +341,109 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+        "-m1",
+        '"Starting an election due to step up request"',
+    )
+
+    procs = []
+    for unit in ops_test.model.applications[APP_NAME].units:
+        kubectl_cmd[5] = unit.name.replace("/", "-")
+        kubectl_proc = subprocess.Popen(kubectl_cmd, stdout=subprocess.PIPE)
+        grep_proc = subprocess.Popen(
+            grep_cmd, stdin=kubectl_proc.stdout, stdout=subprocess.DEVNULL
```
dragomirp, line 371: The election log may drift away, due to `continuous_writes()` spam, so I pipe the logs to make sure it's not missed. Is this a sane approach?

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +341,109 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
         f"{unit.name.replace('/', '-')}.mongodb-k8s-endpoints"
         for unit in ops_test.model.applications[APP_NAME].units
     ]
+
+
+async def db_step_down(ops_test: OpsTest, sigterm_time: datetime) -> None:
+    """Pipes the k8s logs, looking for stepdown message."""
+    kubectl_cmd = [
+        "microk8s",
+        "kubectl",
```
dragomirp, line 350: Should most probably check if we have regular kubectl, so that microk8s isn't a hard dependency for the tests.

tests/integration/ha_tests/helpers.py
MiaAltieri: oh wow this works? It used to be the case that you couldn't run this since it was coming from outside the namespace or something like that. If this works definitely do this.

tests/integration/ha_tests/helpers.py
MiaAltieri: true, but the tests run on a microk8s cloud (i.e. microk8s is already installed) and I don't think we will ever change from that considering Canonicals love for microk8s.

tests/integration/ha_tests/helpers.py
MiaAltieri: YES! Awesome work here, this was a pain for me to deal with as well.

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +341,109 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+    ]
+
+    grep_cmd = (
+        "grep",
+        "-m1",
+        '"Starting an election due to step up request"',
+    )
+
+    procs = []
+    for unit in ops_test.model.applications[APP_NAME].units:
```
MiaAltieri, line 367: comment block for this explaining the election log and piping the logs

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +341,109 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+        kubectl_cmd[5] = unit.name.replace("/", "-")
+        kubectl_proc = subprocess.Popen(kubectl_cmd, stdout=subprocess.PIPE)
+        grep_proc = subprocess.Popen(
+            grep_cmd, stdin=kubectl_proc.stdout, stdout=subprocess.DEVNULL
+        )
+        procs.append((kubectl_proc, grep_proc))
+
+    timeout = 30
+    success = False
+    for kubectl_proc, grep_proc in procs:
```
MiaAltieri, line 377: comment explaining this

tests/integration/ha_tests/test_ha.py
```diff
@@ -65,6 +83,128 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+    # first find primary, write to primary, then read from each unit
+    await insert_focal_to_cluster(ops_test)
+
+    primary = await get_replica_set_primary(ops_test)
+    app = await get_application_name(ops_test, APP_NAME)
+    excluded = [primary]
+    for unit in ops_test.model.applications[app].units:
+        if unit.name == primary:
+            continue
+        with await get_mongo_client(ops_test, excluded=excluded) as client:
```
MiaAltieri, line 169: Make a _new_ helper function that gets a specific mongodb_client for a given unit and add it to helpers.

```
async def get_direct_mongo_client(
    ops_test: OpsTest, unit,
) -> MongoClient:
    """Returns a direct mongodb client for a specified unit."""
    return MongoClient(
       await mongodb_uri(ops_test, [int(unit.name.split("/")[1])]), directConnection=True
    )
```

Then you can just do 
```
    for unit in ops_test.model.applications[app].units:
        with await get_direct_mongo_client(ops_test, unit) as client:
            db = client["new-db"]
            test_collection = db["test_ubuntu_collection"]
            query = test_collection.find({}, {"release_name": 1})
            assert query[0]["release_name"] == "Focal Fossa"
```

tests/integration/ha_tests/test_ha.py
```diff
@@ -65,6 +83,128 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+    with await get_mongo_client(ops_test) as client:
+        actual_writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
+    assert total_expected_writes == actual_writes
+
+
+async def test_replication_member_scaling(ops_test: OpsTest, continuous_writes) -> None:
+    """Verify newly added and newly removed members properly replica data.
+
+    Verify newly members have replicated data and newly removed members are gone without data.
+    """
```
MiaAltieri, line 187: I think you should just add this as a check to the end of `test_scale_down_capablities` and `test_scale_up_capablities` rather than having an additional test here

Maybe you can creat a helper function for:
```

    with await get_mongo_client(ops_test, excluded=original_units) as client:
        db = client["new-db"]
        test_collection = db["test_ubuntu_collection"]
        query = test_collection.find({}, {"release_name": 1})
        assert query[0]["release_name"] == "Focal Fossa"

        # verify that the no writes were skipped
        total_expected_writes = await get_total_writes(ops_test)
        actual_writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
        assert total_expected_writes == actual_writes
```

tests/integration/ha_tests/test_ha.py
MiaAltieri: see above comment about refactoring other tests.

tests/integration/ha_tests/test_ha.py
```diff
@@ -198,18 +332,67 @@ async def test_freeze_db_process(ops_test, continuous_writes):
...
+        old_primary,
+        MONGODB_CONTAINER_NAME,
+        MONGOD_PROCESS_NAME,
+        "SIGTERM",
+    )
+    # verify that a stepdown was performed on restart. SIGTERM should send a graceful restart and
+    # send a replica step down signal. Pipes k8s logs output to see if any of the pods received a
+    # stepdown request. Must be done early otherwise continuous writes may flood the logs
+    await db_step_down(ops_test, sig_term_time)
+
```
MiaAltieri, line 365: Would be good to sleep to wit for re-election

MiaAltieri (changes requested):
Really awesome stuff here! I also LOVE your improvements on the original code as well. Really execellent stuff

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +345,150 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+    for unit in ops_test.model.applications[app].units:
+        if unit.name == primary:
+            role = "Primary"
+        else:
+            role = "Secondary"
+        with await get_direct_mongo_client(ops_test, unit.name) as client:
+            actual_writes = client[TEST_DB][TEST_COLLECTION].count_documents({})
+        assert (
+            total_expected_writes == actual_writes
+        ), f"{role} {unit.name} missed writes to the db."
```
dragomirp, line 493: Most of the tests either check a random node and/or a specific one. I think we can recheck all the nodes and do that consistently across tests.

delgod (approved):

tests/integration/ha_tests/test_ha.py
```diff
@@ -65,6 +84,87 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+
+async def test_replication_across_members(ops_test: OpsTest, continuous_writes) -> None:
+    """Check consistency, ie write to primary, read data from secondaries."""
+    # first find primary, write to primary, then read from each unit
+    await insert_focal_to_cluster(ops_test)
+
+    # verify that the no writes were skipped
+    await find_focal_in_cluster(ops_test)
+    await verify_writes(ops_test)
```
MiaAltieri, line 165: AFAICT `verify_writes` implicitly checks that each replica has the same data. Hence there is no need to also check that focal is replicated across the cluster
```suggestion
    # verify that no writes were missed to the DB and that writes are consistent across all members
    await verify_writes(ops_test)
```

tests/integration/ha_tests/test_ha.py
```diff
@@ -65,6 +84,87 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+
+    # connect to replica set uri and get replica set members
+    member_hosts = await fetch_replica_set_members(ops_test)
+
+    # verify that the replica set members have the correct units
+    assert set(member_hosts) == set(hostnames), "all members not running under the same replset"
+
+    # verify that the no writes were skipped
+    await find_focal_in_cluster(ops_test)
+    await verify_writes(ops_test)
```
MiaAltieri, line 111: see comment on this

MiaAltieri (commented):
One last change - almost there

tests/integration/ha_tests/test_ha.py
dragomirp: They check different DBs. `verify_writes()` checks `db["continuous_writes_database"]["test_collection"]`, while `find_focal_in_cluster()` checks `db["new-db"]["test_ubuntu_collection"]`.

tests/integration/ha_tests/test_ha.py
dragomirp: Deleting the pod doesn't seem to trigger a Juju election.

tests/integration/ha_tests/helpers.py
```diff
@@ -252,8 +259,9 @@ async def send_signal_to_pod_container_process(
     ), f"Failed to send {signal_code} signal, unit={unit_name}, container={container_name}, process={process}"
 
 
-def host_to_unit(host: str) -> str:
-    return "/".join(host.split(".")[0].rsplit("-", 1))
+def host_to_unit(host: str) -> Optional[str]:
+    if host:
+        return "/".join(host.split(".")[0].rsplit("-", 1))
```
shayancanonical, line 264: nit: it's nice to explicitly return `None` if not host. suggest:
```
return "/".join(host.split(".")[0].rsplit("-", 1)) if host else None
```

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +345,150 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+    await stop_writes_action.wait()
+    total_expected_writes = int(stop_writes_action.results["writes"])
+    assert total_expected_writes > 0, "error while getting total writes."
+    return total_expected_writes
+
+
+async def kubectl_delete(ops_test: OpsTest, unit: ops.model.Unit, wait: bool = True) -> None:
+    """Delete the underlying pod for a unit."""
+    kubectl_cmd = (
+        "microk8s",
```
shayancanonical, line 442: this test may fail if run on a non microk8s environment. instead, i would consider using the `kubectl` command only (and running `microk8s config > ~/.kube/config` locally)

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +345,150 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+        "pod",
+        f"--wait={wait}",
+        f"-n{ops_test.model_name}",
+        unit.name.replace("/", "-"),
+    )
+    ret_code, _, _ = await ops_test.run(*kubectl_cmd)
+    assert ret_code == 0, "Unit failed to delete"
+
+
+async def insert_focal_to_cluster(ops_test: OpsTest) -> None:
```
shayancanonical, line 454: consider renaming/refactoring into `insert_record_in_collection`

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +345,150 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+    """Inserts the Focal Fossa data into the MongoDB cluster via primary replica."""
+    primary = await get_replica_set_primary(ops_test)
+    with await get_direct_mongo_client(ops_test, primary) as client:
+        db = client["new-db"]
+        test_collection = db["test_ubuntu_collection"]
+        test_collection.insert_one({"release_name": "Focal Fossa", "version": 20.04, "LTS": True})
+
+
+async def find_focal_in_cluster(ops_test: OpsTest) -> None:
```
shayancanonical, line 463: consider renaming/refactoring into `find_record_in_collection`

tests/integration/ha_tests/helpers.py
```diff
@@ -334,3 +345,150 @@ async def get_units_hostnames(ops_test: OpsTest) -> List[str]:
...
+    """
+    app = await get_application_name(ops_test, APP_NAME)
+    primary = await get_replica_set_primary(ops_test)
+
+    total_expected_writes = await get_total_writes(ops_test)
+    for unit in ops_test.model.applications[app].units:
+        if unit.name == primary:
+            role = "Primary"
+        else:
+            role = "Secondary"
```
shayancanonical, line 488: minor style nit: role = "Primary" if unit.name == primary else "Secondary"

shayancanonical (dismissed):
Looks great! I raised a few minor concerns

tests/integration/ha_tests/test_ha.py
dragomirp: Removing `insert_focal_to_cluster()` and `find_focal_in_cluster()`

tests/integration/ha_tests/test_ha.py
dragomirp: Pod will be recreated and resume being a leader.

MiaAltieri (approved):
AWESOME WORK

shayancanonical (approved):
LGTM! Just pending a response to https://github.com/canonical/mongodb-k8s-operator/pull/71/files#r1037621954

PR #98, opened 2023-01-24
update to jammy

delgod:


tox.ini
```diff
@@ -11,7 +11,7 @@ envlist = fmt, lint, unit
 src_path = {toxinidir}/src/
 tst_path = {toxinidir}/tests/
 lib_path = {toxinidir}/lib/charms/mongodb
-all_path = {[vars]src_path} {[vars]tst_path}
+all_path = {[vars]src_path} {[vars]tst_path} {[vars]lib_path}
```
taurus-forever, line 14: Why do we play this games? The following pass lint without any issues:

```suggestion
all_path = {toxinidir}
```

taurus-forever (approved):
Nice! One minor suggestion.

tox.ini
delgod: we don't want to run checks against fetched 3rd-party libraries

tests/integration/tls_tests/test_tls.py
```diff
@@ -27,11 +27,16 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+        await ops_test.model.deploy(my_charm, num_units=3, resources=resources, series="jammy")
         await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=2000)
 
         config = {"generate-self-signed-certificates": "true", "ca-common-name": "Test CA"}
-        await ops_test.model.deploy(TLS_CERTIFICATES_APP_NAME, channel="beta", config=config)
+        await ops_test.model.deploy(
+            TLS_CERTIFICATES_APP_NAME,
+            channel="beta",
+            config=config,
+            series="focal",
```
MiaAltieri, line 38: I assume TLS-certs is not on jammy yet?

MiaAltieri, line 38: I assume that TLS-certs isn't availbe on jammy yet

MiaAltieri (approved):
groundbreaking 🍯 🐟 (supposed to be jelly-fish)

tests/integration/tls_tests/test_tls.py
delgod: you right, tls-certificates-operator is on focal for now

PR #109, opened 2023-02-17
Cache charmcraft pack container, skip unstable tests except on schedule

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/146

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -63,7 +63,7 @@ pass_env =
 commands =
     poetry install --with integration
     poetry export -f requirements.txt -o requirements.txt
-    poetry run pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,12 +1,93 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        tox-environments:
+          - integration-charm
+          - integration-database-relation
+          - integration-osm-mysql
+          - integration-replication
+          - integration-self-healing
+          - integration-tls
```
delgod, line 51: I think we need rename tests names in tox. currently we test nothing

delgod (changes requested):

MiaAltieri (approved):
looks good to me, curious why that TLS test is failing, I think its related to the issue paolo mentioned last week

delgod (approved):

PR #117, opened 2023-03-01
Cancel CI that is not the latest commit on a branch

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-operator/pull/134

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
   unit-test:
     name: Unit tests
     runs-on: ubuntu-latest
-    timeout-minutes: 15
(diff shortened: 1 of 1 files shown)


MiaAltieri (approved):

taurus-forever:
> Do we need tests for lib/charms/mongodb/v0/mongodb.py in the K8s charm if the VM charm is the owner of that lib (and has unit tests for the lib?)

My $0.05:
 * we **must** test lib in VM (the repo where library is maintained).
 * we **can** test lib in K8s to make sure library behaves as expected in different environment.
 * we **should** test lib in K8s IF library has environment-specific implementations (it will simplify integrations tests troubleshooting).

taurus-forever (approved):

PR #120, opened 2023-03-09
Fix data directory permissions with chown

delgod:
The data directory is not always writable for the "mongodb" user (e.g. on GKE).
Unfortunately, right now, juju/operator-framework doesn't allow set fsGroup and fsGroupChangePolicy via Pod securityContext.
So it is necessary to change data directory ownership via the "chown" cmd.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -46,6 +46,9 @@ MONITOR_PRIVILEGES = {
     "resource": {"db": "", "collection": ""},
     "actions": ["listIndexes", "listCollections", "dbStats", "dbHash", "collStats", "find"],
 }
+DATA_DIR = "/var/lib/mongodb"
+UNIX_USER = "mongodb"
+UNIX_GROUP = "mongodb"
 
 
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -46,6 +46,9 @@
     "resource": {"db": "", "collection": ""},
     "actions": ["listIndexes", "listCollections", "dbStats", "dbHash", "collStats", "find"],
 }
+DATA_DIR = "/var/lib/mongodb"
```
taurus-forever, line 49: Should those files be adopted to reuse one constant?
```
lib/charms/mongodb/v0/helpers.py
36:DATA_DIR = "/var/lib/mongodb"

tests/unit/test_charm.py
59:  "--dbpath=/var/lib/mongodb --auth "
```

taurus-forever (approved):
LGTM, one question.

MiaAltieri (approved):

PR #137, opened 2023-05-05
Expose metrics endpoint + Basic Int Tests

MiaAltieri:
Jira: [DPE-1893](https://warthogs.atlassian.net/browse/DPE-1893)

## Problem
Metrics endpoint not exposed for MongoDB 

## Solution
Expose metrics endpoint similarly to how the VM charm does 

## Int Tests
- Sanity check that endpoints work
- Check endpoints are still functioning after a network cut

## Future PRs
- Add functionality for changing the monitor user password and updating the endpoint correctly so that metrics are still provided

[DPE-1893]: https://warthogs.atlassian.net/browse/DPE-1893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


delgod (approved):

src/charm.py
```diff
@@ -326,7 +355,6 @@ def _pull_licenses(self, container: Container) -> None:
         """Pull licenses from workload."""
         licenses = [
             "snap",
-            "rock",
```
dmitry-ratushnyy, line 329: Minor question - looks like this change is not related to the PR topic. Is this correct?

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+# See LICENSE file for licensing details.
+import time
+import yaml
+
+import ops
+import pytest
+import urllib3
+from pytest_operator.plugin import OpsTest
+from pathlib import Path
+from ..ha_tests import helpers as ha_helpers
```
dmitry-ratushnyy, line 12: is it possible to avoid relative imports?

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+async def test_endpoints_network_cut(ops_test: OpsTest, chaos_mesh):
+    """Verify that endpoint still function correctly after a network cut."""
+    # retrieve a primary unit and a non-primary unit (active-unit). The primary unit will have its
+    # network disrupted, while the active unit allows us to communicate to `mongod`
+    primary = await ha_helpers.get_replica_set_primary(ops_test)
+    active_unit = [
+        unit
+        for unit in ops_test.model.applications[DATABASE_APP_NAME].units
+        if unit.name != primary.name
```
dmitry-ratushnyy, line 78: maybe we can use `filter` here?

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+@pytest.fixture(scope="module")
+def chaos_mesh(ops_test: OpsTest) -> None:
+    ha_helpers.deploy_chaos_mesh(ops_test.model.info.name)
+
+    yield
+
+    ha_helpers.destroy_chaos_mesh(ops_test.model.info.name)
+
+
+async def get_address(ops_test: OpsTest, app_name=DATABASE_APP_NAME, unit_num=0) -> str:
```
dmitry-ratushnyy, line 30: It looks like `app_name` is always equals to `DATABASE_APP_NAME` if so, do we really need it as a param or am I missing something?

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+    address = status["applications"][app_name]["units"][f"{app_name}/{unit_num}"]["address"]
+    return address
+
+
+async def verify_endpoints(ops_test: OpsTest, unit: ops.model.Unit) -> str:
+    """Verifies mongodb endpoint is functional on a given unit."""
+    http = urllib3.PoolManager()
+
+    unit_address = await get_address(ops_test=ops_test)
+    mongodb_exporter_url = f"http://{unit_address}:{MONGODB_EXPORTER_PORT}/metrics"
```
dmitry-ratushnyy, line 42: Maybe we can inline this variable?

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+        resources = {"mongodb-image": METADATA["resources"]["mongodb-image"]["upstream-source"]}
+        await ops_test.model.deploy(my_charm, num_units=3, resources=resources, series="jammy")
+        await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=2000)
+
+
+async def test_endpoints(ops_test: OpsTest):
+    """Sanity check that endpoints are running."""
+    application = ops_test.model.applications[DATABASE_APP_NAME]
+
+    for unit in application.units:
```
dmitry-ratushnyy, line 66: Maybe we can use `asyncio.gather` here? :smile:

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+async def verify_endpoints(ops_test: OpsTest, unit: ops.model.Unit) -> str:
+    """Verifies mongodb endpoint is functional on a given unit."""
+    http = urllib3.PoolManager()
+
+    unit_address = await get_address(ops_test=ops_test)
+    mongodb_exporter_url = f"http://{unit_address}:{MONGODB_EXPORTER_PORT}/metrics"
+    mongo_resp = http.request("GET", mongodb_exporter_url)
+
+    assert mongo_resp.status == 200
```
dmitry-ratushnyy, line 45: Just curious why we don't use `assertEqual` from `unittest` here

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # if configured correctly there should be more than one mongodb metric present
+    mongodb_metrics = mongo_resp._body.decode("utf8")
+    assert mongodb_metrics.count("mongo") > 10
```
dmitry-ratushnyy, line 49: Just curious why we don't use `assertGreater` from `unittest` here

src/charm.py
MiaAltieri: ooh this was a mistake - thank you for noticing 🙏

tests/integration/metrics_tests/test_metrics.py
delgod: It is needed it for one of the HA tests, which deploys two clusters in parallel and doesn't have mixed data.

tests/integration/metrics_tests/test_metrics.py
MiaAltieri: If I had to guess I would say because its an unnecessary import

tests/integration/metrics_tests/test_metrics.py
MiaAltieri: If I had to guess I would say because its an unnecessary import

tests/integration/metrics_tests/test_metrics.py
dmitry-ratushnyy: I think it is better to use `unitests` library when it is possible

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+async def test_endpoints(ops_test: OpsTest):
+    """Sanity check that endpoints are running."""
+    application = ops_test.model.applications[DATABASE_APP_NAME]
+
+    for unit in application.units:
+        await verify_endpoints(ops_test, unit)
+
+
+async def test_endpoints_network_cut(ops_test: OpsTest, chaos_mesh):
```
dmitry-ratushnyy, line 70: looks like `chaos_mesh` is not used in the test

tests/integration/metrics_tests/test_metrics.py
delgod: yes, @dmitry-ratushnyy is right, we don't use `get_address` function in the `test_unique_cluster_dbs` test.

tests/integration/metrics_tests/test_metrics.py
delgod: I think it is needed for `isolate_instance_from_cluster` and `remove_instance_isolation`

tests/integration/metrics_tests/test_metrics.py
delgod: I don't think it will add a significant speed increase, fetch URL should be super fast.

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+async def test_build_and_deploy(ops_test: OpsTest) -> None:
+    """Build and deploy three units of MongoDB and one unit of TLS."""
+    async with ops_test.fast_forward():
+        my_charm = await ops_test.build_charm(".")
+        resources = {"mongodb-image": METADATA["resources"]["mongodb-image"]["upstream-source"]}
+        await ops_test.model.deploy(my_charm, num_units=3, resources=resources, series="jammy")
+        await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=2000)
+
+
+async def test_endpoints(ops_test: OpsTest):
```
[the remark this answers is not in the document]
delgod: I think we can add mypy linter if we want strict typing.

tests/integration/metrics_tests/test_metrics.py
delgod: The generator (current) style has good readability (but inconsistent performance).
I think it is ok to keep it as is.

tests/integration/metrics_tests/test_metrics.py
MiaAltieri: yes this is correct!

tests/integration/metrics_tests/test_metrics.py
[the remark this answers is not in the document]
MiaAltieri: True - feel free to add `mypy` in a follow up PR Dmitry :)

tests/integration/metrics_tests/test_metrics.py
MiaAltieri: What we should be using is the HA helper:
```
mongodb_application_name = await ha_helpers.get_application_name(ops_test, DATABASE_APP_NAME)
```
which retrieves the app name so that we can run these tests on a provided cluster

tests/integration/metrics_tests/test_metrics.py
MiaAltieri: I think this would make things a little less readable. Is this a necessary change for you? If so I can make it

dmitry-ratushnyy (approved):

PR #139, opened 2023-05-16
logs for mongod are present in cos

MiaAltieri:
## Issue
Logs aren't present in COS 

## Solution
1. pipe logs to file
2. make loki aware of the logs

## Testing 
Screenshot 2023-05-16 at 14 12 20


tests/unit/test_charm.py
```diff
@@ -60,7 +60,8 @@ def test_mongod_pebble_ready(self, connect_exporter, fix_data_dir, defer, pull_l
                     "command": (
                         "mongod --bind_ip_all "
                         "--replSet=mongodb-k8s "
-                        f"--dbpath={DATA_DIR} --auth "
+                        f"--dbpath={DATA_DIR} "
+                        "--logpath=/var/lib/mongodb/mongodb.log --auth "
```
dmitry-ratushnyy, line 64: maybe we can use constants here as well? 😄

delgod (approved):

PR #140, opened 2023-05-17
fix(typo): Rename eporter to exporter

merkata:
I was reading the code and just noticed there's a typo in a comment. This fixes it and reassures devs that it is good to use comments - people do read them.

tl;dr s/eporter/exporter/


MiaAltieri (approved):
Thank you for this contribution @merkata we really appreciate it :)

In the future, we might need to update our tox file so that it includes spelling checks in the linter.

delgod (approved):

PR #150, opened 2023-06-02
[DPE-2048] Stabilize HA tests

dmitry-ratushnyy:
* Fix the HA tests by replacing  a command to get MongoDB start arguments.
 "ps -o cmd fp" command id replaced to  "cat /proc/{pid}/cmdline" because the
former one was truncating long output string.

* Update bash script to cleanup chaos mesh used in HA testing
* Add scripts for cleanup resources.

### DIFF
--- a/tests/integration/cleanup_resources.py
+++ b/tests/integration/cleanup_resources.py
@@ -6,10 +6,11 @@ import argparse
 import json
 import os
 import subprocess
-
+import logging
+logger = logging.getLogger(__name__)
 
 def cleanup_chaos_mesh(namespace) -> None:
-    print(f"Cleaning up chaos mesh in {namespace}")
+    logger.info(f"Cleaning up chaos mesh in {namespace}")
     env = os.environ
     env["KUBECONFIG"] = os.path.expanduser("~/.kube/config")
 
@@ -18,16 +19,16 @@ def cleanup_chaos_mesh(namespace) -> None:
         shell=True,
         env=env,
     )
+    logger.info("Destroy chaos mesh output:")
     for line in output.decode("utf-8").splitlines():
-        print(line)
+        logger.info(line)
 
 
 def cleanup_juju_models() -> None:
     for model in _filter_tests_models("admin/test-"):
-        print(f"Destroying model {model}")
+        logger.info(f"Destroying model {model}")
         delete_cmd = ["juju", "destroy-model", model, "--destroy-storage", "-y"]
-        subprocess.check_output(delete_cmd).decode("utf-8")
-
+        subprocess.check_output(delete_cmd)
 
 def cleanup_all(namespace: str) -> None:
     cleanup_chaos_mesh(namespace)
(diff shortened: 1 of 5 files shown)


tests/integration/ha_tests/helpers.py
```diff
@@ -259,7 +259,7 @@ async def get_process_pid(
         return_code == 0
     ), f"Failed getting pid, unit={unit_name}, container={container_name}, process={process}"
 
-    stripped_pid = pid.strip()
+    stripped_pid = pid.strip().replace("\r", "").replace("\n", "")
```
delgod, line 262:
```suggestion
    stripped_pid = pid.strip().replace("\r", " ").replace("\n", " ")
```
I think in such case “--” replace will not be needed below

dmitry-ratushnyy: This `mongod_cmd.replace("--", " --")` is needed because `cat /proc/{pid}/cmdline` returns this string
```
mongod--bind_ip_all--replSet=mongodb-k8s--dbpath=/var/lib/mongodb--logpath=/var/lib/mongodb/mongodb.log--auth--clusterAuthMode=keyFile--keyFile=/etc/mongod/keyFile
```

dmitry-ratushnyy: This change was maid to prevent exceptions on int(stripped_pid)

delgod (approved):

tests/integration/cleanup_resources.py
```diff
@@ -0,0 +1,63 @@
...
+    env = os.environ
+    env["KUBECONFIG"] = os.path.expanduser("~/.kube/config")
+
+    output = subprocess.check_output(
+        f"tests/integration/ha_tests/scripts/destroy_chaos_mesh.sh {namespace}",
+        shell=True,
+        env=env,
+    )
+    for line in output.decode("utf-8").splitlines():
+        print(line)
```
Mehdi-Bendriss, line 22: Would be better to replace this with a logger

tests/integration/cleanup_resources.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def cleanup_juju_models() -> None:
+    for model in _filter_tests_models("admin/test-"):
+        print(f"Destroying model {model}")
```
Mehdi-Bendriss, line 27: Same comment (logger)

tests/integration/cleanup_resources.py
```diff
@@ -0,0 +1,63 @@
...
+        delete_cmd = ["juju", "destroy-model", model, "--destroy-storage", "-y"]
+        subprocess.check_output(delete_cmd).decode("utf-8")
```
Mehdi-Bendriss, line 29: I don't think you need to `decode("utf-8")` since you don't use the output

Mehdi-Bendriss (approved):

tests/integration/cleanup_resources.py
dmitry-ratushnyy: Done, thank you

tests/integration/cleanup_resources.py
dmitry-ratushnyy: Done, thank you

tests/integration/cleanup_resources.py
dmitry-ratushnyy: Done, thank you

PR #156, opened 2023-06-08
Fix scale down logic

dmitry-ratushnyy:
The goal of this PR is to fix scale down logic of mongodb unit.

This is achieved by blocking unit removal until the unit is not removed from replica set configuration

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -11,6 +11,7 @@ includes scaling and other capabilities.
 """
 
 import logging
+import time
 from typing import Dict, Optional
 
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
@@ -38,7 +39,7 @@ from charms.mongodb.v0.mongodb import (
 from charms.mongodb.v0.mongodb_provider import MongoDBProvider
 from charms.mongodb.v0.mongodb_tls import MongoDBTLS
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
-from ops.charm import ActionEvent, CharmBase
+from ops.charm import ActionEvent, CharmBase, RelationDepartedEvent
 from ops.main import main
 from ops.model import ActiveStatus, Container
 from ops.pebble import ExecError, Layer, PathError, ProtocolError
@@ -55,6 +56,7 @@ UNIX_USER = "mongodb"
 UNIX_GROUP = "mongodb"
 MONGODB_EXPORTER_PORT = 9216
 REL_NAME = "database"
+ON_STOP_TIMEOUT = 1000
 
 
 class MongoDBCharm(CharmBase):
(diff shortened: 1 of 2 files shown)


delgod (approved):

src/charm.py
```diff
@@ -214,16 +217,44 @@ def _on_start(self, event) -> None:
...
+        with MongoDBConnection(self.mongodb_config) as mongo:
+            try:
+                replset_members = mongo.get_replset_members()
+                return self.get_hostname_by_unit(self.unit.name) in replset_members
+            except NotReadyError as e:
+                logger.error(f"{self.unit.name}.is_unit_in_replica_set NotReadyError={e}")
+            except PyMongoError as e:
+                logger.error(f"{self.unit.name}.is_unit_in_replica_set PyMongoError={e}")
+
+    def _on_stop(self, event) -> None:
```
Mehdi-Bendriss, line 231: I believe you could move this logic to the `storage_detaching` hook which is the first hook to be called on scaledown. And you already have some logic related to the scaledown there.

And do not use the `stop` hook (following conversation with @marcoppenheimer about upgrades, this hook gets called on a juju upgrade action - so better leave it for now)

src/charm.py
```diff
@@ -214,16 +217,44 @@ def _on_start(self, event) -> None:
...
+        if "True" == self.unit_peer_data.get("unit_departed", "False"):
+            logger.debug(f"{self.unit.name} blocking on_stop")
+            is_in_replica_set = True
+            timeout = ON_STOP_TIMEOUT
```
Mehdi-Bendriss, line 235: nit: you could use `tenacity.retry` to abstract the retry logic

tests/integration/test_teardown.py
```diff
@@ -49,31 +49,9 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
-        for count in scales:
-            await scale_and_verify(ops_test, count=count)
-
-
-@pytest.mark.skip("Issues with juju teardown")
-@pytest.mark.abort_on_fail
-async def test_scale_up_scale_down_large(ops_test: OpsTest):
-    """Scale up and down the application and verify the replica set is healthy."""
-    scales = [6, -6, 7, -7]
+    scales = [2, -1, -1, 2, -2, 3, -3, 4, -4, 5, -5, 6, -6, 7, -7]
```
Mehdi-Bendriss, line 54: Nice

src/charm.py
dmitry-ratushnyy: I am not sure about using `storage_detaching`, the usage of `stop` hook was suggested by juju devs. The idea is to block deletion of unit and the `stop` is the correct place to do it

src/charm.py
dmitry-ratushnyy: I don't think that using `renacity` here is the god idea

src/charm.py
Mehdi-Bendriss: `stop` comes later in the teardown process, `storage_detaching` is the entry point of teardown (you can confirm this by using `jhack utils tail -l 30 -d` then deleting the unit and watch the lifecycle) so it doesn't hurt to move the logic there imo. For reference: https://github.com/canonical/mongodb-operator/blob/main/src/charm.py#L155

dmitry-ratushnyy: we still need to be sure what by the time app leader receives `RelationDepartedEvent` for a unit, the unit  and its pod is not deleted.

dmitry-ratushnyy:
Closing as obsolete

PR #164, opened 2023-06-20
[DPE-2156] Rename keys for user's secrets to be compatible with juju 3.1

dmitry-ratushnyy:
# About
The goal of the PR is to update user related secrets key names to be compatible with juju 3.1

# Related issues
DPE-2156


juditnovak (approved):
LGTM :-)

PR #167, opened 2023-06-23
[MISC] First commit of Documentation

deusebio:


docs/reference/r-overview.md
deusebio, line 1: I have to remove the link from the top level in the navigation table, since the sync between github and discourse would not work otherwise. This is a temporary fix, as we may address this in the sync tool

deusebio:
superseded by #199

PR #176, opened 2023-07-05
[DPE-2078] Fix scaled down for Juju 3.1.5

dmitry-ratushnyy:
## About

The goal of this pull request is to fix scale down issue for Juju 3.1.5. 
To do that we are moving logic to remove unit from a cluster from `on_storage_detaching` to `on_stop` event.

## Related issue
DPE-2078

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -4,6 +4,7 @@
 # See LICENSE file for licensing details.
 
 import logging
+import time
 from typing import List, Optional, Set
 
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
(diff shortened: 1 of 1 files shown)


delgod (approved):

src/charm.py
```diff
@@ -295,6 +292,10 @@ def _relation_changes_handler(self, event) -> None:
         """Handles different relation events and updates MongoDB replica set."""
         self._connect_mongodb_exporter()
 
+        if type(event) is RelationDepartedEvent:
```
Mehdi-Bendriss, line 295: Interesting approach

Mehdi-Bendriss (approved):

PR #186, opened 2023-08-01
[DPE-2320] rework secret id comparison

delgod:
Using `secret.get_info().id` instead of `secret.id` always returns a short secret id.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -4,7 +4,6 @@
 # See LICENSE file for licensing details.
 
 import logging
-import re
 import time
 from typing import Dict, List, Optional, Set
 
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -564,12 +543,12 @@ def _on_secret_changed(self, event):
...
-        if self._compare_secret_ids(
-            event.secret.id, self.app_peer_data.get(Config.Secrets.SECRET_INTERNAL_LABEL)
+        if event.secret.get_info().id == self.app_peer_data.get(
+            Config.Secrets.SECRET_INTERNAL_LABEL
         ):
             scope = APP_SCOPE
-        elif self._compare_secret_ids(
-            event.secret.id, self.unit_peer_data.get(Config.Secrets.SECRET_INTERNAL_LABEL)
+        elif event.secret.get_info().id == self.unit_peer_data.get(
+            Config.Secrets.SECRET_INTERNAL_LABEL
```
juditnovak, line 551: I think we should use the URI here... ID may overlap across charms, while URI (that's BTW including the ID) never would. What do U think?

delgod:
we are going to use labels

PR #80, opened 2022-12-02
Adding full cluster crash/restart test

dragomirp:
# Cluster crash test failures:
Cluster crash/restart currently fail the assertion for total writes, meaning that the DB didn't persist some of the writes `continuous_writes.py` thinks it should have.

The cause for the mismatch seems to be a PyMongo OperationError thrown while restarted nodes rejoin the cluster with the following message:
```
No keys found for HMAC that is valid for time: { ts: Timestamp(1669922958, 10) } with id: 7172123513443057669, full error: {'ok': 0.0, 'errmsg': 'No keys found for HMAC that is valid for time: { ts: Timestamp(1669922958, 10) } with id: 7172123513443057669', 'code': 211, 'codeName': 'KeyNotFound'}
```

## Steps to reproduce manually
* Deploy a cluster of three mongodb-k8s units
* Deploy the HA test charm from /tests/integration/ha_tests/application_charm
* Create a layer configuration to prevent pebble from restarting the service and upload it to the MongoDB units in the mongod container:
```
services:
  mongod:
    override: merge
    on-success: ignore
    on-failure: ignore
```
* Add the layer to pebble and replan:
```
/charm/bin/pebble add test2 <your_layer_config_location>
/charm/bin/pebble replan
```
* Kill all mongod processes in all the units
* Replan the services to restart them and check the output of `continuous_writes.py`. You should get errors about the time the second or third node rejoins.

## Potentially related log output from mongod logs:
```
2022-12-01T11:32:55.643Z [mongod] {"t":{"$date":"2022-12-01T11:32:55.643+00:00"},"s":"I",  "c":"-",
  "id":4939300, "ctx":"monitoring-keys-for-HMAC","msg":"Failed to refresh key cache","attr":{"error":"ReadConcernMajorityNotAvailableYet: Read concern majority reads are currently not possible.","nextWakeupMillis":600}}
```
```
2022-12-01T13:37:22.043Z [mongod] {"t":{"$date":"2022-12-01T13:37:22.042+00:00"},"s":"I",  "c":"CONTROL",  "id":20714,   "ctx":"LogicalSessionCacheRefresh","msg":"Failed to refresh session cache, will try again at the next refresh interval","attr":{"error":"NotYetInitialized: Replication has not yet been configured"}}
```

## Potentially related issues:
* [Same error, but seems to be for older MongoDB versions](https://jira.mongodb.org/browse/SERVER-47568)
[...]

PR #190, opened 2023-08-23
[DPE-2412]  Full cluster crash test

MiaAltieri:
## Problem
Current HA test suite doesnt have a full cluster crash test

## Solution
Add full cluster crash test

## Reused code
Code from the Postgres repo that changes the pebble layer was ported over (i.e. all the changes in `helpers`)

## Additional Context
to get this test to function properly, some modifications were needed in continuous writes. Specifically we must try again when the HMAC is out of sync between the client and the cluster. The reasoning for this was discussed in Mattermost but is pasted below:

I'm running into the issue that @dragop ran into when he first attempted to do [full cluster crash test](https://github.com/canonical/mongodb-k8s-operator/pull/80) which is:

> Cluster crash/restart currently fail the assertion for total writes, meaning that the DB didn't persist some of the writes `continuous_writes.py` thinks it should have.
The cause for the mismatch seems to be a PyMongo OperationError thrown while restarted nodes rejoin the cluster with the following message:
```
No keys found for HMAC that is valid for time: { ts: Timestamp(1669922958, 10) } with id: 7172123513443057669, full error: {'ok': 0.0, 'errmsg': 'No keys found for HMAC that is valid for time: { ts: Timestamp(1669922958, 10) } with id: 7172123513443057669', 'code': 211, 'codeName': 'KeyNotFound'}
```

After some googling, I *think* that vector clock cluster time is out of sync, between the replicas and the client. (This would make sense since we shutdown the `mongod` process on all replicas and I assume the client needs to get insync again). From Wiki:
> In [cryptography](https://en.wikipedia.org/wiki/Cryptography "Cryptography"), an **HMAC** (sometimes expanded as either **keyed-hash message authentication code** or **hash-based message authentication code**) is a specific type of [message authentication code](https://en.wikipedia.org/wiki/Message_authentication_code "Message authentication code") (MAC) involving a cryptographic hash function and a secret cryptographic key. As with any MAC, it may be used to simultaneously verify both the data integrity and authenticity of a message.

According to the [MongoDB Docs](https://www.mongodb.com/docs/v5.0/core/security-client-side-encryption/#encryption-algorithms) , HMAC encryption is used on the *client side*, so my theory is that the client that we use to write to MongoDB has out of date HMAC keys, (due to the cluster going down and then being available again). 

From what I can tell in my testing by hand is that we miss only one write to the DB, so I assume that the client only *temporarily* has out of date HMAC keys, and that is resolves this and writes continue. After all, the test does show that writes continue to the DB after the cluster comes back up. 

I believe *This is NOT an issue with the MongoDB cluster. The correct way to handle this is on the client side.**  i.e. when we encounter an HMAC error the client should try again. 

more reading shows this error in the logs on the cluster side:
```
  "id":4939300, "ctx":"monitoring-keys-for-HMAC","msg":"Failed to refresh key cache","attr":{"error":"ReadConcernMajorityNotAvailableYet: Read concern majority reads are currently not possible.","nextWakeupMillis":600}}
```
so I think that when the replica set slowly comes back up and gets reconfigured, there is a moment where the HMAC key is out of date on the replica set side. **AFAICT this is reasonable and to be expected, since the replica set needs to get back in sync. I don't think we can expect the replica set to restart and immediately have the correct HMAC key** 

This reasoning was then verified by Mykola

### DIFF
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -7,6 +7,8 @@ import os
 import string
 import subprocess
 import tempfile
+import tarfile
+import kubernetes as kubernetes
 from asyncio import gather
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


tests/integration/ha_tests/helpers.py
```diff
@@ -730,3 +732,158 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
         )
         ret_code, _, _ = await ops_test.juju(*replan_cmd)
         assert ret_code == 0, f"Failed to replan for unit {unit.name}"
+
+
+async def are_all_db_processes_down(ops_test: OpsTest, process: str) -> bool:
+    """Verifies that all units of the charm do not have the DB process running."""
+    app = await get_application_name(ops_test, APP_NAME)
+
+    if "/" in process:
```
dmitry-ratushnyy, line 741: is it possible to add a comment here on what exactly this check does or maybe create a with an explanatory name ?

MiaAltieri: sure thing!

dmitry-ratushnyy (dismissed):

dmitry-ratushnyy (dismissed):

tests/integration/ha_tests/helpers.py
```diff
@@ -730,3 +732,160 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+                ), f"Failed to replan pebble layer, unit={unit_name}, container={container_name}, service={service_name}"
+
+
+def copy_file_into_pod(
+    client: kubernetes.client.api.core_v1_api.CoreV1Api,
+    namespace: str,
+    pod_name: str,
+    container_name: str,
+    destination_path: str,
+    source_path: str,
```
deusebio, line 844: **super niptick** Maybe it is more intuitive to swap destination and source, to make it consistent to what we are used to `cp source target`, using shell

tests/integration/ha_tests/helpers.py
```diff
@@ -730,3 +732,160 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+        for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+            with attempt:
+                for unit in ops_test.model.applications[app].units:
+                    _, raw_pid, _ = await ops_test.juju(
+                        "ssh", "--container", MONGODB_CONTAINER_NAME, unit.name, *pgrep_cmd
+                    )
+
+                    # If something was returned, there is a running process.
+                    if len(raw_pid) > 0:
+                        raise ProcessRunningError
```
deusebio, line 758: **question** if this fails for all attempts, does the `Retrying` raise a `RetryError`? Does the `ProcessRunningError` ever bubble up or it is always masked by the RetryError?

tests/integration/ha_tests/helpers.py
```diff
@@ -730,3 +732,160 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+    container_name = "mongod"
+    service_name = "mongod"
+    now = datetime.now().isoformat()
+
+    copy_file_into_pod(
+        client,
+        ops_test.model.info.name,
+        pod_name,
+        container_name,
+        f"/tmp/pebble_plan_{now}.yml",
```
deusebio, line 792: **niptick** as this string is used at least twice,maybe it is worth creating a variable to centralize its value

tests/integration/ha_tests/manifests/extend_pebble_restart_delay.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  mongod:
+    override: merge
+    backoff-delay: 150s
+    backoff-limit: 150s
```
deusebio, line 5: **niptick** shouldn't this be 180s to convert to 3min, as mentioned in the comments?

deusebio (approved):
LGTM

deusebio (approved):

dratushnyy (approved):

dmitry-ratushnyy (approved):

PR #194, opened 2023-08-30
[DPE-2414] Storage reuse

MiaAltieri:
## Problem
Current HA Test Suite does not test whether MongoDB properly re-uses storage. 

## Context
What it means to reuse storage is that when storage is reattached to the juju unit, that the mongo daemon recognises this storage and instead of performing the typical sync from the primary that it instead only syncs what is necessary by computing the diff of what the old storage already has

In [Juju versions before 2.9.43](https://chat.charmhub.io/charmhub/pl/rm1ub9wmw3dr3fzetf1fgieezc), this cannot be tested as there was a bug with storage reuse

## Solution
Add a test that tests storage reuse


tests/integration/ha_tests/helpers.py
```diff
@@ -734,6 +735,58 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+        "cat /var/lib/mongodb/mongodb.log",
+    ]
+
+    return_code, start_states, _ = await ops_test.juju(*cat_cmd)
+
+    assert (
+        return_code == 0
+    ), f"Failed catting mongodb logs, unit={reused_unit.name}, container={MONGODB_CONTAINER_NAME}"
+
+    for line in start_states.split("\n"):
```
dmitry-ratushnyy, line 761: nits: you can use .filter() here

tests/integration/ha_tests/helpers.py
```diff
@@ -734,6 +735,58 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+        if not len(line):
+            continue
+
+        item = json.loads(line)
+
+        re_use_time = convert_time(item["t"]["$date"])
```
dmitry-ratushnyy, line 767: could you please change variable name to "reuse_time" because the first association with "re_*" is regex

tests/integration/ha_tests/test_ha.py
```diff
@@ -586,3 +588,48 @@ async def test_network_cut(ops_test: OpsTest, continuous_writes, chaos_mesh):
...
+    )
+
+    # k8s will automatically use the old storage from the storage pool
+    removal_time = datetime.now(timezone.utc).timestamp()
+    await scale_application(ops_test, app, current_number_units)
+    await ops_test.model.wait_for_idle(
+        apps=[app], status="active", timeout=1000, wait_for_exact_units=(current_number_units)
+    )
+
+    new_unit = get_newest_unit(ops_test, app)
```
deusebio, line 621: **suggstion** maybe we could do a further assertion here: we could get the new_unit before scaling down, and after scaling up, and assert that the unit has the same ordinal number. I'm not sure there is a property to check that they have a different hash or something (since they are two different pods...), but if there is, I would check for this

deusebio (commented):
Uhm, I'm not sure how the tests can pass given the typo, or maybe I have misunderstood the code :shrug: 

Apart from that, code looks all right! Just some small comment and a suggestion

dmitry-ratushnyy (dismissed):

tests/integration/ha_tests/helpers.py
```diff
@@ -734,6 +735,60 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+        "cat /var/lib/mongodb/mongodb.log",
+    ]
+
+    return_code, logs, _ = await ops_test.juju(*cat_cmd)
+
+    assert (
+        return_code == 0
+    ), f"Failed catting mongodb logs, unit={reused_unit.name}, container={MONGODB_CONTAINER_NAME}"
+
+    filtered_logs = filter(filter_logs.split("\n"), logs)
```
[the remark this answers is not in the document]
MiaAltieri: oops sorry - you're correct. I think I pushed too quickly after addressing the last round of comments. my bad

tests/integration/ha_tests/helpers.py
```diff
@@ -734,6 +735,60 @@ async def update_pebble_plans(ops_test: OpsTest, override: Dict[str, str]) -> No
...
+
+
+def convert_time(time_as_str: str) -> int:
+    """Converts a string time representation to an integer time representation."""
+    # parse time representation, provided in this format: 'YYYY-MM-DDTHH:MM:SS.MMM+00:00'
+    d = datetime.strptime(time_as_str, "%Y-%m-%dT%H:%M:%S.%f%z")
+    return time.mktime(d.timetuple())
+
+
+def get_newest_unit(ops_test: OpsTest, app_name: str) -> Unit:
```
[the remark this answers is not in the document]
MiaAltieri: good point! Thank you Enrico!

tests/integration/ha_tests/test_ha.py
MiaAltieri: I like this suggestion!  But for the purpose of this test, I believe this function is sufficient and would like to keep it simple. (i.e. we want to get the most recently added unit from juju, if we were to scale the app by more than one unit then I think there might be some inconsistencies with which unit is newer.) I'll add a comment to the code to make this more clear. Thanks again Enrico :)

dmitry-ratushnyy (approved):

deusebio (approved):
LGTM

PR #200, opened 2023-09-20
[WIP][DPE-2447] Secret labels

juditnovak:


.github/workflows/ci.yaml
```diff
@@ -14,31 +14,31 @@ on:
...
-    runs-on: ubuntu-latest
-    timeout-minutes: 15 # TODO: reduce to 5 minutes
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-      - name: Install tox
-        # TODO: Consider replacing with custom image on self-hosted runner OR pinning version
-        run: python3 -m pip install tox
-      - name: Run tests
-        run: tox run -e unit
```
juditnovak, line 41: This PR should be rebased on `main` that is ONLY using Juju3.

The ugly commented out stuff should "natively" disappear after the rebase (conflicts resolved, sry :sweat_smile: )

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
...
+from ops.model import SecretNotFoundError
+
+from config import Config
+from exceptions import SecretAlreadyExistsError
+
+# The unique Charmhub library identifier, never change it
+
+# FIXME: Incorrect libid, as my charmcraft authentication is missing 'package access' :-/
+LIBID = "b9a7fe0c38d8486a9d1ce94c27d4758e"
+
```
juditnovak, line 18: Must be fixed -- I got blocked on permission issues :-(

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
...
+APP_SCOPE = Config.Relations.APP_SCOPE
+UNIT_SCOPE = Config.Relations.UNIT_SCOPE
+Scopes = Config.Relations.Scopes
+
+
+def generate_secret_label(charm: CharmBase, scope: Scopes) -> str:
+    """Generate unique group_mappings for secrets within a relation context."""
+    members = [charm.app.name, scope]
+    return f"{'.'.join(members)}"
+
```
juditnovak, line 35: I really think this function should be left separate from the `SecretCache` class(es).

**Rationale:**

`SecretCache`: Providing an abstraction layer above direct Juju access WITH caching. I.e. making sure not to fetch a secret multiple times. (*)
Now, if at some point we decide we want caching no more, this is the single, encapsulated logic that is to be changed (dropped). `SecretChache` and `CachedSecret` are to provide this level of isolation.

`generate_secret_label`: The logic how to define it's secret labesl **belongs ot the application**. Each charm can decide how to label their secrets. This is a different separation of concern either from Juju access or from internal implementation details whether we're cacnihg secrets or not. This is part of the Application Logic.

With all the above in mind, pls in case someone changing the PR -- if possible pls keep `generate_secrets_label()` separate from the abstract `Secret`-manipulation classes. This function is outside of that scope.

(*) For multiple reasons actullay, including that the secret may have changed. If we're not carefully passing around our already-fetched secret (when dealing with a chain of called functions), but fetch it again, we should get the latest version... Which could be different from what we had at the start of the event handler ;-)

lib/charms/mongodb/v0/mongodb_secrets.py
juditnovak: Done

delgod (dismissed):

lib/charms/mongodb/v0/mongodb_secrets.py
MiaAltieri: Yes, this info in the doc strings would be awesome! Thank you so much for taking time to add all this info.

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
...
+        self._secret_uri = secret_uri
+        self.label = label
+        self.charm = charm
+
+    def add_secret(self, content: Dict[str, str], scope: Scopes) -> Secret:
+        """Create a new secret."""
+        if self._secret_uri:
+            raise SecretAlreadyExistsError(
+                "Secret is already defined with uri %s", self._secret_uri
+            )
```
MiaAltieri, line 58: noob question what is a `secret_uri` is it just the mongodb URI?

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
...
+            if not (self._secret_uri or self.label):
+                return
+            try:
+                self._secret_meta = self.charm.model.get_secret(label=self.label)
+            except SecretNotFoundError:
+                if self._secret_uri:
+                    self._secret_meta = self.charm.model.get_secret(
+                        id=self._secret_uri, label=self.label
+                    )
+        return self._secret_meta
```
MiaAltieri, line 81: nit
```suggestion
        if self._secret_meta:
            return self._secret_meta
            
        if not (self._secret_uri or self.label):
            return
        try:
            self._secret_meta = self.charm.model.get_secret(label=self.label)
        except SecretNotFoundError:
            if self._secret_uri:
                self._secret_meta = self.charm.model.get_secret(
                    id=self._secret_uri, label=self.label
                )
       return self._secret_meta
```

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def add_secret(self, content: Dict[str, str], scope: Scopes) -> Secret:
+        """Create a new secret."""
+        if self._secret_uri:
+            raise SecretAlreadyExistsError(
+                "Secret is already defined with uri %s", self._secret_uri
+            )
+
+        if scope == Config.APP_SCOPE:
+            secret = self.charm.app.add_secret(content, label=self.label)
```
MiaAltieri, line 61: neat, what a cool function `app.add_secret`! This code looks fine, but to help my understanding can you answer these two questions:

1. From what I can tell *any* unit can call this function correct? 
2. And there aren't any concerns about multiple units calling this function right?

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
...
+    def get_content(self) -> Dict[str, str]:
+        """Getting cached secret content."""
+        if not self._secret_content:
+            if self.meta:
+                self._secret_content = self.meta.get_content()
+        return self._secret_content
+
+    def set_content(self, content: Dict[str, str]) -> None:
+        """Setting cached secret content."""
+        if self.meta:
```
MiaAltieri, line 92: I am a noob, is it worth it to have only the leader execute this or does it not matter?

src/charm.py
```diff
@@ -687,8 +686,8 @@ def _set_user_created(self, user: MongoDBUser) -> None:
     def _get_mongodb_config_for_user(
         self, user: MongoDBUser, hosts: List[str]
     ) -> MongoDBConfiguration:
-        external_ca, _ = self.tls.get_tls_files(UNIT_SCOPE)
-        internal_ca, _ = self.tls.get_tls_files(APP_SCOPE)
+        external_ca, _ = self.tls.get_tls_files(APP_SCOPE)
+        internal_ca, _ = self.tls.get_tls_files(UNIT_SCOPE)
```
MiaAltieri, line 690: I am pretty sure this change should be reverted
```suggestion
        external_ca, _ = self.tls.get_tls_files(UNIT_SCOPE)
        internal_ca, _ = self.tls.get_tls_files(APP_SCOPE)
```

Context: 
- Internal secrets for TLS are used for the MongoDB replicas to communicate with each other with an encrpyted connection, for this, we need all units to have the SAME internal TLS secrets. Since all units should have the **the same** internal secret, it is crucial that this secret is kept consistent across all units. To ensure this, only allow one unit to change this data, so we use "lock" for changing this field, this "lock" is granted to the leader unit. This prevents multiple units from changing the internal secret. APP_DATA works perfect for this because APP_DATA only allowed to be changed by the leader.

- external secrets are used for each unit to communicate outside of the cluster and do not need to be the same for each unit, hence writing to unit data is best here

dmitry-ratushnyy (approved):

src/charm.py
```diff
@@ -1177,5 +1103,8 @@ def _set_data_dir_permissions(container: Container) -> None:
     # END: static methods
 
 
+# END: Secret cache
+
+
```
MiaAltieri, line 1108: should this be here? It looks like there are other non-secret functions before this. We also have no `START` section. 

suggestions you can either:
1. move it to the correct place and add a `START` section
2. remove it

src/charm.py
```diff
@@ -804,120 +803,47 @@ def _set_leader_unit_active_if_needed(self):
...
+        if not secret:
+            return
 
-        return self.secrets[scope][Config.Secrets.SECRET_LABEL].id
+        value = secret.get_content().get(key)
+        if value != Config.Secrets.SECRET_DELETED_LABEL:
+            return value
 
     def set_secret(self, scope: Scopes, key: str, value: Optional[str]) -> Optional[str]:
         """(Re)defining a secret."""
```
MiaAltieri, line 818: please forgive my noob question, but why isn't it necessary to check if the unit is a leader

tests/integration/test_charm.py
```diff
@@ -149,6 +149,16 @@ async def test_reset_and_get_password_secret_same_as_cli(ops_test: OpsTest) -> N
     # Getting back the pw programmatically
     password = await get_password(ops_test, unit_id=leader_id, username="monitor")
 
+    #
+    # Noway to retrieve a secet by label for now (https://bugs.launchpad.net/juju/+bug/2037104)
```
MiaAltieri, line 153: nit `No way`

lib/charms/mongodb/v0/mongodb_secrets.py
```diff
@@ -0,0 +1,125 @@
+"""Secrets related helper classes/functions."""
```
MiaAltieri, line 1: This is an excellent file. Just so you know we have the VM Charm for MongoDB own the shared libraries, so in our development we try to create/change libraries on the MongoDB VM charm and only ever retrieve libraries on the MongoDB K8s charm. What is more is that I am fairly certain we do not allow publishing of libraries on the CI of K8s charms. 

With that said, I am personally okay with you making this change here since you have already written this excellent code! But I ask that after this PR merges that:
1. You create a follow up PR that adds this library to the MongoDB VM charm (I will immediately approve if the PR has **only this library**)
2. In the future if you would like to make changes to this library that you make them on the VM charm

MiaAltieri (changes requested):
Looks AMAZING Judit, thank you so much for your dedication and help here. 

Mostly of my comments are questions; but there are a few nits and a requested change

src/charm.py
juditnovak: GRRRRRR indeed, VERY GOOD POINT, THANK YOU!!!!!!!!!!!!!!!!!!
I'm also wondering that some tests should have shown an error on this :-( :-( :-( 
(I did this change in desperation when trying to resolve a horrible issue coming from a different source. I can't believe I left it in the code THANK U SO MUCH 4 NOTICING!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!)

lib/charms/mongodb/v0/mongodb_secrets.py
juditnovak: That sounds ALL wonderful.... 
I would like to do these changes equally on the VM as well anyway ;-)

lib/charms/mongodb/v0/mongodb_secrets.py
juditnovak: Nononono (not that I knew what 'mongodb uri' means :sweat_smile: )

Secret URI is basically the model-qualified identifier of each secret. Something like:

 - `secret://<model-UUID>/<secrect-XID>`
 
 where the XID is a unique identifier of each secret

juditnovak: Nononono (not that I knew what 'mongodb uri' means :sweat_smile: )

Secret URI is basically the model-qualified identifier of each secret. Something like:

 - `secret://<model-UUID>/<secrect-XID>`
 
 where the XID is a unique identifier of each secret

lib/charms/mongodb/v0/mongodb_secrets.py
juditnovak: Any unit may have their own secrets (typicaly unit-level TLS certs).

The logic (that U may be referring to) is that `app` level secrets shouldn't be modified by any other unit but the lead.

Which is true indeed -- however, in my opinion that logic should be handled outside of this scope. 
(And it actually is.)

This module to me should be "dumb" in this sense, Focusing on secrets handling (caching + communication with the Juju Secret store), and no additional business logic (incl. who may or may not be allowed to change a secret).

lib/charms/mongodb/v0/mongodb_secrets.py
juditnovak: Nononono, we have unit-level secrets too!!!!

src/charm.py
juditnovak: As this question is coming up a lot, I thought to bring up a live example :-)

https://github.com/canonical/mongodb-k8s-operator/blob/main/lib/charms/mongodb/v0/mongodb_tls.py#L177

Typically TLS certs exist both on the `app` level (controlled by the leader unit) and on an individual level for units themselves. Thus units must be able to set their own secrets -- without being a leader.

src/charm.py
juditnovak: Bah, seems like a leftover as the `mongodb_secrets.py` (new lib) conents were here originally, thank U for noticing!!!

lib/charms/mongodb/v0/mongodb_secrets.py
MiaAltieri: Ah thank you for teaching me!

lib/charms/mongodb/v0/mongodb_secrets.py
MiaAltieri: Thank you for the context 🙇

MiaAltieri (approved):
Really awesome work Judit, thank you very much!

dmitry-ratushnyy (approved):

PR #202, opened 2023-09-25
[DPE-2610] Remove support of Juju 2.9

dmitry-ratushnyy:
The goal of this PR is to remove support of Juju 2.9 from the charm.

To do so we are removing code related to handling secrets for juju 2.9


MiaAltieri (changes requested):
looks good, but dont forget to add
```
          juju-channel: 3.1/stable
          bootstrap-options: "--agent-version 3.1.5"
```
to ci or the tests will fail. [you can use vm CI as a reference](https://github.com/canonical/mongodb-operator/blob/6/edge/.github/workflows/ci.yaml)

pyproject.toml
```diff
@@ -45,7 +45,7 @@ lightkube = "^0.13.0"
 pytest = "^7.3.1"
 pytest-mock = "^3.11.1"
 pytest-operator = "^0.27.0"
-juju = "2.9.44.0 || 3.2.0.1"
+juju = "3.2.0.1"
```
delgod, line 48: maybe, can be separate PR
```suggestion
juju = "3.2.2"
```

tox.ini
```diff
@@ -70,7 +70,7 @@ pass_env =
     CI
 commands_pre =
     poetry install --with integration
-    poetry run pip install juju==2.9.44.0
+    poetry run pip install juju==3.2.0.1
```
delgod, line 73: same as above
```suggestion
    poetry run pip install juju==3.2.2
```

delgod (approved):

pyproject.toml
dmitry-ratushnyy: Will commit in this one  if you don't mind

MiaAltieri (dismissed):

delgod (approved):

PR #206, opened 2023-10-10
[DPE-2734] Extended TLS testing

juditnovak:
This attempt relates to [an issue on a previous PR](https://github.com/canonical/mongodb-k8s-operator/pull/200#discussion_r1350066961).... that REALLY shouldn’t have happened.

The mistake was major, and no tests alerted about it.

In case next time Mia may be biking around in the Himalayas – we should have an automated way to prevent such heavy errors :grin:


tests/integration/tls_tests/test_tls.py
```diff
@@ -84,6 +147,8 @@ async def test_rotate_tls_key(ops_test: OpsTest) -> None:
         new_internal_cert_time = await time_file_created(ops_test, unit.name, INTERNAL_CERT_PATH)
         new_mongod_service_time = await time_process_started(ops_test, unit.name, DB_SERVICE)
 
+        check_certs_correctly_distributed(ops_test, unit)
```
juditnovak, line 150: Hmm, I'm also thinking to add this to the following test, `test_set_tls_key()`, where we're setting specific keys.

tests/integration/helpers.py
```diff
@@ -257,3 +257,81 @@ async def secondary_mongo_uris_with_sync_delay(ops_test: OpsTest, rs_status_data
...
+            f"no relation data could be grabbed on relation with endpoint {relation_name} and alias {relation_alias}"
+        )
+
+    return relation_data[0]["application-data"].get(key)
+
+
+async def get_secret_id(ops_test, app_or_unit: Optional[str] = None) -> str:
+    """Retrieve secert ID for an app or unit."""
+    complete_command = "list-secrets"
+    # if app_or_unit:
```
dmitry-ratushnyy, line 321: Is there a reason why it is commented?

MiaAltieri (approved):
This looks good to me, but conceptually I think this sort of testing would be best for a unit test

tests/integration/helpers.py
juditnovak: Yeah... It doesn't work ;-)
I'm already discussing it with the Juju team (and manually filter in this code instead).

I'm finalizing this part after it's clear whether the Juju command could do the job, or manual filter remains.

dmitry-ratushnyy: so should we wait for the juju team response? I would prefer not to have commented code in the tests :)

juditnovak: Juju team have already responded, I just didn't yet have the time to check their response.

Leaving comments in the code has never been the plan. But either to remove them, or to use them replacing the at-the-time working solution.

[Suggestion](https://chat.charmhub.io/charmhub/pl/6wqy9hbfcbb5ufs6mm9bkfrz3e) from the Juju team worked, the code is adjusted accordingly.

dmitry-ratushnyy (dismissed):

MiaAltieri (dismissed):

MiaAltieri (approved):

dmitry-ratushnyy (approved):

dmitry-ratushnyy (approved):

PR #207, opened 2023-10-16
[DPE-2632] Updated data platform libs to use external secrets

dmitry-ratushnyy:
# About

The goal of this PR is to update data platform libs to use external secrets in MongoDB charm and update the charm code and tests if required


delgod (dismissed):

lib/charms/mongodb/v0/mongodb_provider.py
```diff
@@ -29,7 +29,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
```
MiaAltieri, line 31: Question: It looks like you made these changes yourself, since they don't match what you have in your branch on [the VM charm](https://github.com/canonical/mongodb-operator/blob/932c7f69ec96358b8ca5beae66756b2ef9a9addc/lib/charms/mongodb/v0/mongodb_provider.py)

Typically on K8s charms we should be doing `charmcraft fetch-libs` so that libraries don't get out of sync. Can you wait until your VM code merges, then copy that code over?

MiaAltieri (changes requested):

juditnovak (approved):
LGTM

lib/charms/mongodb/v0/mongodb_provider.py
dmitry-ratushnyy: I did fetch lib on both charms today morning after a new version of data platform libs was published to charmhub 😄

MiaAltieri (approved):

PR #221, opened 2023-11-20
Fix branch used to trigger release to 5/edge

dmitry-ratushnyy:
## About
The goal of this PR is to fix the branch that will trigger publishing to 5/edge channel

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -360,6 +360,7 @@ class MongoDBCharm(CharmBase):
         It is needed to install mongodb-clients inside the charm container
         to make this function work correctly.
         """
+
         container = self.unit.get_container(Config.CONTAINER_NAME)
         if not container.can_connect():
             logger.debug("mongod container is not ready yet.")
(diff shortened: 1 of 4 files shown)


delgod (approved):

MiaAltieri (approved):

src/charm.py
```diff
@@ -590,6 +606,42 @@ def _on_secret_changed(self, event):
...
 
     # BEGIN: user management
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(5),
+        reraise=True,
+        before=before_log(logger, logging.DEBUG),
+    )
+    def _initialise_users(self, event: StartEvent) -> None:
+        """Create users."""
```
MiaAltieri, line 616:
```suggestion
        """Create users.
        
        User creation can only be completed after the replica set has been initialised which requires some time. In race conditions this can lead to failure to initialise users. To prevent these race conditions from breaking the code, retry on failure.
        """
```

MiaAltieri (changes requested):

delgod (approved):

MiaAltieri (approved):

PR #226, opened 2023-11-22
Split cluster initialisation into two parts: cluster initialisation and user creation

dmitry-ratushnyy:
## About
Sometimes MongoDB need more time to initialise replica set an this can lead to a race condition and failure on user creation.

## Solution
Split `_initialise_replica_set` function into two parts:
1. Cluster initialisation
2. User creation

and add retries for user creation


MiaAltieri (approved):
Thank you!!

delgod (approved):

Issue #212, opened 2023-10-31
Failed to get pbm status errors


gruyaume:
## Steps to reproduce
1. juju deploy mongodb-k8s
2. juju deploy sdcore-nrf
3. juju integrate mongodb-k8s sdcore-nrf

## Expected behavior
No errors

## Actual behavior

Failed to get pbm status errors in charm.

## Versions

Operating system: Ubuntu 22.04
Juju CLI: 3.1.6
Juju agent: 3.1.6
Charm revision: 6/edge (Revision 37)
microk8s: v1.27.6

## Log output
```
unit-mongodb-k8s-0: 07:37:32 ERROR unit.mongodb-k8s/0.juju-log Failed to get pbm status. non-zero exit code 1 executing ['/usr/bin/pbm', 'status', '-o', 'json'], stdout='{"Error":"get status of pitr: check for errors: get current epoch: get config: get: mongo: no documents in result"}\n', stderr=''
```

MiaAltieri:
Hi @gruyaume from what I can tell this is infact an issue in the log messages. I'm happy you noticed it since poor logging can impact how users understand the charm, thanks for pointing this out :) 

The log messages come from the shared mongodb libraries (which are hosted & maintained on the VM side) right now there are two open PRs to fix this: [6/edge](https://github.com/canonical/mongodb-operator/pull/305) and [5/edge](https://github.com/canonical/mongodb-operator/pull/306) once these are merged we can update the libraries on the k8s side.

Other than that though, the charm should be functional. Did you encounter any functional errors or only the incorrect log?

gruyaume:
Perfect, thank you @MiaAltieri

PR #227, opened 2023-11-22
fix logging (6/edge) - via copying solution over from VM charm

MiaAltieri:
## Issue
Backup status logs error message, when not related to S3. Instead this should be info

## Context
Reported as a bug in #212 

## Solution
Copy changes made from VM [see Merged pr](https://github.com/canonical/mongodb-operator/pull/305)

### DIFF
--- a/lib/charms/mongodb/v0/mongodb_backups.py
+++ b/lib/charms/mongodb/v0/mongodb_backups.py
@@ -13,7 +13,7 @@ import logging
 import re
 import subprocess
 import time
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from charms.data_platform_libs.v0.s3 import CredentialsChangedEvent, S3Requirer
 from charms.mongodb.v0.helpers import (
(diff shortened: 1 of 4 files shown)


delgod (dismissed):

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -35,14 +35,14 @@
 )
 
 # The unique Charmhub library identifier, never change it
-LIBID = "9f2b91c6128d48d6ba22724bf365da3b"
+LIBID = "18c461132b824ace91af0d7abe85f40e"
```
dmitry-ratushnyy, line 38: Is there a reason why LIBID updated?

MiaAltieri: yes, old file had the wrong LIBID

dmitry-ratushnyy (approved):

PR #233, opened 2023-12-05
[DPE-3111] Fix integration tests for Charmed MongoDB 6

dmitry-ratushnyy:
## About

The goal of this pull request is to stabilise integration tests. 

## How

Instead of creating a helper pod every time when we need to run a command in MongoDB cluster, create the pod once per test to avoid receiving an empty output from a command.


.github/workflows/ci.yaml
```diff
@@ -108,14 +108,23 @@ jobs:
...
+      #   run: sg snap_microk8s -c "tox run -e ${{ matrix.tox-environments }} -- --model testing -m '${{ steps.select-tests.outputs.mark_expression }}'"
+      #   env:
+      #     CI_PACKED_CHARMS: ${{ needs.build.outputs.charms }}
+
+
+      # - name: Dump logs
+      #   uses: canonical/charm-logdump-action@main
+      #   if: failure()
+      #   with:
+      #     app: mongodb-k8s
```
MiaAltieri, line 129: wait I am a noob, I think this fixes + disables them.

Shouldn't it only fix them? Or did I miss something during the call today

tests/integration/helpers.py
```diff
@@ -135,18 +138,23 @@ async def run_mongo_op(
...
 
     logger.info("Running mongo command: %r", mongo_cmd)
+
+    create_pod_if_not_exists(
+        ops_test.model_name, HELPER_MONGO_POD_NAME, "mongo", f"mongo:{HELPER_MONGO_VERSION}"
+    )
+
+    while not is_pod_ready(ops_test.model_name, HELPER_MONGO_POD_NAME):
+        logger.info("Waiting for pod to be ready...")
+        time.sleep(5)
```
MiaAltieri, line 148: oh epic, thank you for this. I think it will also be necessary to port over this fix to k8s charm. Mind also doing that?

MiaAltieri (changes requested):
looking awesome

.github/workflows/ci.yaml
dmitry-ratushnyy: This is done on purpose for debug tests on CI (which I am doing at the moment) :)

tests/integration/helpers.py
dmitry-ratushnyy: you mean VM charm 😄 ? Yes, I am planning to do that

delgod (approved):

src/charm.py
```diff
@@ -387,6 +413,11 @@ def _on_start(self, event) -> None:
             return
 
         self._initialise_replica_set(event)
+        try:
+            self._initialise_users(event)
+        except RetryError:
+            event.defer()
```
MiaAltieri, line 419: logger message might be nice here

tests/integration/ha_tests/test_ha.py
```diff
@@ -559,28 +559,35 @@ async def test_network_cut(ops_test: OpsTest, continuous_writes, chaos_mesh):
...
-    new_primary = await get_replica_set_primary(ops_test, excluded=[primary.name])
+    new_primary = await get_replica_set_primary(
+        ops_test, excluded=[primary.name], use_subprocess_to_get_password=True
+    )
     assert new_primary.name != primary.name
 
     # Remove networkchaos policy isolating instance from cluster
     remove_instance_isolation(ops_test)
 
+    time.sleep(60)
```
MiaAltieri, line 585: might be nice to add a comment so we know what the sleep is for

MiaAltieri (approved):

src/charm.py
dmitry-ratushnyy: Added log message

tests/integration/ha_tests/test_ha.py
dmitry-ratushnyy: Done, thank you

MiaAltieri (approved):

delgod (approved):

PR #239, opened 2024-01-19
[DPE-2910] Enable audit logging

dmitry-ratushnyy:
## About

The goal of this PR is to enable audit logging for Charmed MongoDB K8s and add audit logs to CoS

## Audit log
Screenshot 2024-01-19 at 16 45 30

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -46,6 +46,8 @@ containers:
     mounts:
       - storage: mongodb
         location: /var/lib/mongodb
+      - storage: mongodb-logs
+        location: /var/log/mongodb
 resources:
   mongodb-image:
     type: oci-image
@@ -55,4 +57,6 @@ storage:
   mongodb:
     type: filesystem
     location: /var/lib/mongodb
-  
+  mongodb-logs:
+    type: filesystem
+    location: /var/log/mongodb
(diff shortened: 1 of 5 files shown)


metadata.yaml
```diff
@@ -46,6 +46,8 @@ containers:
     mounts:
       - storage: mongodb
         location: /var/lib/mongodb
+      - storage: mongodb-logs
+        location: /var/log/mongodb
```
delgod, line 50: I think (please check) adding storage will break "juju refresh".
Could we add logs to the data dir right now? and decide if it should be separate dir next few weeks

delgod (changes requested):

metadata.yaml
dmitry-ratushnyy: sure, let's add logs for data dir first and then adjust when necessary. Same needs to be done for VM charm anyway

Mehdi-Bendriss (dismissed):

delgod (dismissed):

MiaAltieri (approved):
looks good, but you'll **definitely** want those metrics tests to be passing before you merge

delgod (approved):

Mehdi-Bendriss (approved):

PR #244, opened 2024-04-19
[DPE-4189] Fix TLS double-restart + update charm components - move to DP lib

Mehdi-Bendriss:
## Issue
- The `tls-certificate-operator` and tls v1 interface is deprecated
- we're using many old versions of python dependencies and charm libs 
- multiple parts of the TLS integ. tests are not asserted / awaited / correct, effectively not validating the behavior of TLS
- the CI/CD pipelines don't use DP workflows
- every single new issue of certificate (internal / external) triggers a restart - during the same flow

## Solution
- migrate to the DP workflows 
- removes dependencies to the deprecated `tls-certificate-operator` + v1 interface
- introduces the `self-signed-certificates operator`  + v3 interface
- upgrades charm libs 
- upgrade python dependencies 
- upgrade dp action libs to the latest
- fix  tests:
    - CSR retrievals from secrets
    - CA cert retrieval from units
    - missing `assert`s and `await`s on coroutines
 - fix additional (double) TLS restart

This PR updates all dependencies of the charm:
- migration to dp workflows
- CI / CD workflows
- python dependencies 
- charm libs 
- TLS operator references


.github/workflows/sync_issue_to_jira.yaml
```diff
@@ -9,13 +9,13 @@ on:
 jobs:
   sync:
     name: Sync GitHub issue to Jira
-    uses: canonical/data-platform-workflows/.github/workflows/sync_issue_to_jira.yaml@v8
+    uses: canonical/data-platform-workflows/.github/workflows/sync_issue_to_jira.yaml@v13.1.2
     with:
       jira-base-url: https://warthogs.atlassian.net
       jira-project-key: DPE
-      jira-component-names: mongodb-k8s
+      jira-component-names: postgresql-k8s
```
MiaAltieri: heads up here

tests/integration/tls_tests/test_tls.py
```diff
@@ -255,7 +260,7 @@ async def test_set_tls_key(ops_test: OpsTest) -> None:
         new_internal_cert_time = await time_file_created(ops_test, unit.name, INTERNAL_CERT_PATH)
         new_mongod_service_time = await time_process_started(ops_test, unit.name, DB_SERVICE)
 
-        check_certs_correctly_distributed(ops_test, unit)
+        await check_certs_correctly_distributed(ops_test, unit)
```
MiaAltieri, line 263: @Mehdi-Bendriss does this work? IIRC there were extensive changes needed for this to work on VM

MiaAltieri:
@Mehdi-Bendriss how did you update the libs, I assume with fetch lib? Its complaining about the libs having local changes but I just wanted to check

tests/integration/tls_tests/test_tls.py
Mehdi-Bendriss: I think (hope?) It should as there is no sharding involved only internal vs external TLS resources

.github/workflows/sync_issue_to_jira.yaml
Mehdi-Bendriss: Oops thanks for the catch!

Mehdi-Bendriss:
For the failed check, I'm not entirely sure what to do with it tbh. As I had to copy past the new `mongodb_tls`  lib from the mongodb VM charm, as it wasn't published on charmhub...

Mehdi-Bendriss:
@carlcsaposs-canonical - could you review the part of this PR where we're moving to the DP worfklows? 
(diff is huge, but mostly libs updates)

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
 name: Tests
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: "${{ github.workflow }}-${{ github.ref }}"
```
carlcsaposs-canonical, line 6: why this change?

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
   cancel-in-progress: true
 
 on:
   pull_request:
+    paths-ignore:
+      - '.gitignore'
+      - '.jujuignore'
+      - 'LICENSE'
+      - '**.md'
+      - .github/renovate.json5
```
carlcsaposs-canonical: imo, the few times that this will reduce a CI run isn't worth the UX confusion from sometimes running ci.yaml and sometimes not running ci.yaml

(I've ran into situations where I was trying to trigger ci by updating e.g. a readme file and then debugging why ci didn't run)

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
+        uses: actions/checkout@v4
+      - name: Install tox & poetry
+        run: |
+          pipx install tox
+          pipx install poetry
       - name: Run tests
         run: tox run -e unit
 
   build:
+    name: Build "${{ matrix.path }}" charm
```
carlcsaposs-canonical:
```suggestion
    name: Build charm | ${{ matrix.path }}
```
nit: follow existing convention

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v13.1.2
     strategy:
       fail-fast: true
```
carlcsaposs-canonical: fail-fast is default, no need to add

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
+        run: |
+          pipx install tox
+          pipx install poetry
       - name: Run tests
         run: tox run -e unit
 
   build:
+    name: Build "${{ matrix.path }}" charm
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v13.1.2
     strategy:
```
carlcsaposs-canonical, line 58: nit (convention): put strategy as first key (before `name`)

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
       fail-fast: true
       matrix:
-        charms: [".", "tests/integration/ha_tests/application_charm", "tests/integration/relation_tests/application-charm"]
-    name: Build ${{matrix.charms}} charm
-    uses: canonical/data-platform-workflows/.github/workflows/build_charm_without_cache.yaml@v8
+        path:
+          - "."
+          - "tests/integration/ha_tests/application_charm"
+          - "tests/integration/relation_tests/application-charm"
```
carlcsaposs-canonical: nit
```suggestion
          - .
          - tests/integration/ha_tests/application_charm
          - tests/integration/relation_tests/application-charm
```

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
     with:
-      charmcraft-snap-revision: 1349
-      path-to-charm-directory: ${{matrix.charms}}
+      cache: true
+      path-to-charm-directory: "${{ matrix.path }}"
```
carlcsaposs-canonical:
```suggestion
      path-to-charm-directory: ${{ matrix.path }}
```

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
-    strategy:
-      fail-fast: false
-      matrix:
-        tox-environments:
-          - charm-integration
-          - ha-integration
-          - tls-integration
-          - metrics-integration
-    name: ${{ matrix.tox-environments }}
+    name: Integration test charm | 3.1.7
```
carlcsaposs-canonical, line 69:
```suggestion
    name: Integration test charm
```
nit: no need to include version since you're only testing on one juju version; also don't want to have another place where version needs to be updated if changed (that could get out of sync & provide incorrect info)

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
-        with:
-          app: mongodb-k8s
-          model: testing
+    uses: canonical/data-platform-workflows/.github/workflows/integration_test_charm.yaml@v13.1.2
+    with:
+      artifact-prefix: packed-charm-cache-true
+      cloud: microk8s
+      microk8s-snap-channel: 1.29-strict/stable
+      juju-agent-version: 3.1.7
+      libjuju-version-constraint: 3.4.0.0
```
carlcsaposs-canonical:
```suggestion
```
libjuju should be pinned in your poetry.lock

see https://github.com/canonical/data-platform-workflows/blob/f86cfdfbc92c929928c0722e7542867db0b092cd/.github/workflows/integration_test_charm.yaml#L48-L58

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
+    secrets:
+      integration-test: |
+        {
+          "AWS_ACCESS_KEY": "${{ secrets.AWS_ACCESS_KEY }}",
+          "AWS_SECRET_KEY": "${{ secrets.AWS_SECRET_KEY }}",
+          "GCP_ACCESS_KEY": "${{ secrets.GCP_ACCESS_KEY }}",
+          "GCP_SECRET_KEY": "${{ secrets.GCP_SECRET_KEY }}",
+        }
```
carlcsaposs-canonical:
```suggestion
        {  "AWS_ACCESS_KEY": "${{ secrets.AWS_ACCESS_KEY }}",
          "AWS_SECRET_KEY": "${{ secrets.AWS_SECRET_KEY }}",
          "GCP_ACCESS_KEY": "${{ secrets.GCP_ACCESS_KEY }}",
          "GCP_SECRET_KEY": "${{ secrets.GCP_SECRET_KEY }}", }
```
(https://github.com/canonical/data-platform-workflows/pull/166)

.github/workflows/ci.yaml
```diff
@@ -3,120 +3,95 @@
...
+    permissions:
+      contents: write  # Needed for Allure Report beta
```
carlcsaposs-canonical:
```suggestion
```
allure not enabled

.github/workflows/release.yaml
```diff
@@ -30,16 +30,14 @@ jobs:
...
   release:
       name: Release charm
       needs:
         - ci-tests
         - build
-      uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v8
+      uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v13.1.2
       with:
         channel: 6/edge
         artifact-name: ${{ needs.build.outputs.artifact-name }}
```
carlcsaposs-canonical: incorrect parameter & output

.github/workflows/sync_issue_to_jira.yaml
```diff
@@ -18,4 +18,4 @@ jobs:
       jira-api-token: ${{ secrets.JIRA_API_TOKEN }}
       jira-user-email: ${{ secrets.JIRA_USER_EMAIL }}
     permissions:
-      issues: write  # Needed to create GitHub issue comment
+      issues: write  # Needed to create GitHub issue comment
```
carlcsaposs-canonical, line 21: nit: newline

pyproject.toml
```diff
@@ -14,43 +14,65 @@ homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
...
-jsonschema = "^4.17.3"
+python = "^3.10.12"
+ops = "^2.12.0"
+pymongo = "^4.6.3"
+tenacity = "^8.2.3"
+cryptography = "^42.0.5"
+jsonschema = "^4.21.1"
 pyyaml = "^6.0.1"
-cosl = "0.0.7"
+cosl = "^0.0.11"
```
carlcsaposs-canonical: is `cosl` a direct dep or from a lib?

pyproject.toml
```diff
@@ -14,43 +14,65 @@ homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
...
-[tool.poetry.group.fmt.dependencies]
-black = "^23.3.0"
-isort = "^5.12.0"
+[tool.poetry.group.charm-libs.dependencies]
+ops = "^2.12.0"
+pydantic = "^1.10.7"
+cryptography = "^42.0.5"
+jsonschema = "^4.21.1"
+cosl = "^0.0.11"
+bcrypt = "^4.0.1"
```
carlcsaposs-canonical: would suggest adding a comment where each dep comes from & also using the same version pinning as the lib (so `*` if unpinned by the lib)

e.g. https://github.com/canonical/mysql-router-operator/blob/15c2d7eeca1012f47e3a4858f2a680e8745726d2/pyproject.toml#L20-L29

pyproject.toml
```diff
@@ -14,43 +14,65 @@ homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
...
+pytest-operator = "^0.34.0"
+parameterized = "^0.9.0"
+cryptography = "^42.0.5"
+jsonschema = "^4.21.1"
+cosl = "^0.0.11"
+bcrypt = "^4.0.1"
+juju = "3.4.0"
+pytest-operator-cache = {git = "https://github.com/canonical/data-platform-workflows", tag = "v13.1.1", subdirectory = "python/pytest_plugins/pytest_operator_cache"}
+pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", tag = "v13.1.1", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
+pytest-github-secrets = {git = "https://github.com/canonical/data-platform-workflows", tag = "v13.1.1", subdirectory = "python/pytest_plugins/github_secrets"}
```
carlcsaposs-canonical: nit: please put pytest-github-secrets after pytest

and the other two after pytest-operator

pyproject.toml
```diff
@@ -14,43 +14,65 @@ homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
...
+pytest = "^8.1.1"
+pytest-asyncio = "^0.21.1"
+pytest-mock = "^3.14.0"
+pytest-operator = "^0.34.0"
+parameterized = "^0.9.0"
+cryptography = "^42.0.5"
+jsonschema = "^4.21.1"
+cosl = "^0.0.11"
+bcrypt = "^4.0.1"
+juju = "3.4.0"
```
carlcsaposs-canonical:
```suggestion
juju = "^3.4.0"
```

pyproject.toml
```diff
@@ -14,43 +14,65 @@ homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
...
+pymongo = "^4.6.3"
+lightkube = "^0.15.2"
+pytest = "^8.1.1"
+pytest-asyncio = "^0.21.1"
+pytest-mock = "^3.14.0"
+pytest-operator = "^0.34.0"
+parameterized = "^0.9.0"
+cryptography = "^42.0.5"
+jsonschema = "^4.21.1"
+cosl = "^0.0.11"
```
carlcsaposs-canonical: are these direct deps in your integration tests?

requirements.txt
carlcsaposs-canonical, line 1: requirements.txt should not be version controlled

tox.ini
```diff
@@ -13,141 +13,82 @@ lib_path = {tox_root}/lib/charms/mongodb
...
-    PYTHONPATH = {tox_root}/lib:{[vars]src_path}
-    PYTHONBREAKPOINT=ipdb.set_trace
-    PYTHONDONTWRITEBYTECODE=1
-    PY_COLORS=1
-pass_env =
-    PYTHONPATH
-    CHARM_BUILD_DIR
-    MODEL_SETTINGS
-deps =
+    PYTHONPATH = {tox_root}:{tox_root}/lib:{[vars]src_path}
```
carlcsaposs-canonical: tox_root should not be on PYTHONPATH

tox.ini
```diff
@@ -13,141 +13,82 @@ lib_path = {tox_root}/lib/charms/mongodb
...
-commands =
-    poetry run pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_charm.py
-
-[testenv:ha-integration]
-description = Run HA integration tests
-pass_env =
-    {[testenv]pass_env}
     CI
+    GITHUB_OUTPUT
+    S3_INTEGRATOR_CHARMPATH
```
carlcsaposs-canonical: what is this var needed for?

tox.ini
```diff
@@ -13,141 +13,82 @@ lib_path = {tox_root}/lib/charms/mongodb
...
-    poetry run pytest -v --tb native --log-cli-level=INFO -s --durations=0 {posargs} {[vars]tests_path}/integration/backup_tests/test_backups.py
-
-[testenv:cleanup_chaos_mesh]
-description = Cleanup chaos mesh
-commands =
-     python {[vars]tests_path}/integration/cleanup_resources.py --cleanup "cleanup_chaos_mesh" --namespace "{posargs}" 
-
-[testenv:cleanup_juju_models]
-description = Cleanup Juju models
+    poetry install --only main,charm-libs,integration
```
carlcsaposs-canonical:
```suggestion
    poetry install --only integration
```

pyproject.toml
carlcsaposs-canonical, line 1: nit: add optional deps groups following https://github.com/canonical/mysql-router-k8s-operator/blob/main/pyproject.toml so that `poetry install` creates the correct virtual env for an IDE venv

carlcsaposs-canonical (commented):
(only reviewed data-platform-workflows related/adjacent changes)

.github/workflows/ci.yaml
Mehdi-Bendriss: Intellij linting raises an error otherwise

carlcsaposs-canonical: I think that's because it evaluates the yaml before ${{ }} variables are replaced by github (when workflow runs)

I believe it's an issue in the linter, not in the yaml

.github/workflows/ci.yaml
Mehdi-Bendriss: I set this one so it's clear to us what juju version is used, to remind us to upgrade/bump if needed

.github/workflows/ci.yaml
Mehdi-Bendriss: It is already the case - I wanted to be verbose to clearly show "juju agent version, libjuju, microk8s" versions (so we don't leave things forgotten and not update them like before) 
Is there a risk to adding it here as well?

.github/workflows/ci.yaml
Mehdi-Bendriss: Can we enable it?

pyproject.toml
Mehdi-Bendriss: It's a dep of a charm lib

pyproject.toml
Mehdi-Bendriss: What is the reasoning here? 
I prefer these 3 grouped at the end of the list, as they expand on the whole width of the line and that interrupts reading coming from short named dependencies

pyproject.toml
Mehdi-Bendriss: Done, but I suspect I'll have to revisit this later for the `*` part

pyproject.toml
Mehdi-Bendriss: Not sure I'm following, seems like we have the same list of groups?

tox.ini
Mehdi-Bendriss: Not sure - copy/past from opensearch

.github/workflows/ci.yaml
Mehdi-Bendriss: I think it doesn't really make a difference in the end evaluated YAML, in which case I'd prefer having no errors in my IDE

.github/workflows/ci.yaml
Mehdi-Bendriss: I would argue it's more readable and efficient to have the `name` as a first entry point to a job or a step

pyproject.toml
Mehdi-Bendriss: no, they're direct dependencies of charm libs used in the integ. tests

.github/workflows/ci.yaml
carlcsaposs-canonical: (we have a renovate config for that)

at least, I would encourage you to consider putting it in a variable so that it correctly reflects whichever arg is passed/version is actually used

.github/workflows/ci.yaml
carlcsaposs-canonical: please read the description I linked—you should not use the parameter unless you need to override the constraint in your pyproject.toml—specifying this will actively override your locked dependencies

> I wanted to be verbose

extending that logic, every python dependency should be pinned here—this is not what this input is for

.github/workflows/ci.yaml
carlcsaposs-canonical: if you accept breaking changes on patch or minor version bump, yes

see here: https://github.com/canonical/data-platform-workflows/pull/131

pyproject.toml
carlcsaposs-canonical: then it should go in charm-libs group, not main dep group

pyproject.toml
carlcsaposs-canonical: they're plugins to the dependencies that I mentioned

pytest-github-secrets hooks into pytest
pytest-operator-* hooks into pytest-operator

pyproject.toml
carlcsaposs-canonical: see `optional`, e.g. https://github.com/canonical/mysql-router-k8s-operator/blob/f9c2fdb27a1e599bf972c705ae2f48280b94009f/pyproject.toml#L30-L31

.github/workflows/ci.yaml
carlcsaposs-canonical: name is built from matrix

regardless, the current convention in data-platform-workflows and our other repos it to put the matrix first (imo, it improves readability a lot to be able to quickly identify if a job uses a matrix, since it affects how you debug it)

for the sake of consistency, please consider following the existing convention (also, if you ask me for help with debugging your workflows in the future, it will make my life easier)

pyproject.toml
carlcsaposs-canonical: do your integration tests import from charm libs? if so, just opensearch charm libs or do they also import libs that use `cosl`? confused about what `cosl` would be used for integration test code

carlcsaposs-canonical: just trying to point out dependencies that might not be intentional. if intentional, please ignore

Mehdi-Bendriss: thank you!

.github/workflows/ci.yaml
Mehdi-Bendriss: Added a variable for it

tests/integration/tls_tests/test_tls.py
MiaAltieri: When we co-worked on the [fix TLS Pr for VM](https://github.com/canonical/mongodb-operator/pull/395/files#diff-9f99a813697ed7cb55d5b5fa50b90b7f37c8c532c2c9fbac05b4b87ab5a11e36) the extensive changes for tests to pass were also applied to the non-sharding tests.

We can separate it out into a separate PR and I can fix it there, but to keep the commit history on this repo good, I would kindly request that you change this PR name to `[DPE-4189] update charm components - move to DP `

MiaAltieri (dismissed):

.gitignore
```diff
@@ -13,3 +13,6 @@ pyvenv.cfg
 share/
 .idea/
 .tox/
+
+requirements.txt
+requirements-last-build.txt
```
carlcsaposs-canonical:
```suggestion
/requirements.txt
/requirements-last-build.txt
```
nit: if you have other charms (e.g. test charms) in this repo, I imagine you'd use requirements.txt instead of poetry for those charms (no need for dep locking imo) so you'd probably want to version control those requirements.txt files

pyproject.toml
```diff
@@ -5,52 +5,73 @@
...
-[tool.poetry.group.fmt.dependencies]
-black = "^23.3.0"
-isort = "^5.12.0"
+[tool.poetry.group.charm-libs.dependencies]
+ops = "^2.12.0"
+pydantic = "^1.10.7"  # cos_agent lib
+cryptography = "^42.0.5"  # tls_certificates lib v3
+jsonschema = "^4.22.0"  # tls_certificates lib v3
+cosl = "^0.0.11"  # loki_push_api
+bcrypt = "^4.0.1"
```
carlcsaposs-canonical: same as https://github.com/canonical/mongodb-k8s-operator/pull/244#discussion_r1586007623 about avoiding adding extra constraint beyond the lib's constraint

MiaAltieri (approved):

.github/workflows/ci.yaml
carlcsaposs-canonical: nit: ^

.github/workflows/sync_issue_to_jira.yaml
carlcsaposs-canonical: nit: ^

pyproject.toml
carlcsaposs-canonical: ^

.github/workflows/ci.yaml
carlcsaposs-canonical: nit: ^

carlcsaposs-canonical (commented):
(nits are non-blocking; 1 non-nit comment)

pyproject.toml
Mehdi-Bendriss: Bcrypt was already removed in a previous commit, or am I mistaken?

Mehdi-Bendriss:
Thanks @carlcsaposs-canonical - I believe all the non-nit feedback has been addressed (?) I did not address the nits I did not agree with

carlcsaposs-canonical:
> I believe all the non-nit feedback has been addressed

are the additional version constraints on charm lib dependencies intentional? (https://github.com/canonical/mongodb-k8s-operator/pull/244#discussion_r1590719699)

Mehdi-Bendriss:
Yes.

carlcsaposs-canonical (approved):

PR #245, opened 2024-04-23
move user init code out of repl set init code

MiaAltieri:
## Issue
CI tests are failing to create users 

On further inspection [of the debug log in this workflow](https://github.com/canonical/mongodb-k8s-operator/actions/runs/8797992063/job/24145056569?pr=243), in the function `_initialise_replica_set`, there are still calls to initialise users.

## Solution
remove any calls to initialise users in `_initialise_replica_set`

## Other
Also the fact that `_initialise_replica_set` sets `db_initialised` to true is problematic. Because if the `_initialise_users` functions fails, then another hooks will check whether `db_initialised` is true and proceed to fail since the user they will try to access the DB with won't have been created yet

### DIFF
--- a/lib/charms/mongodb/v0/mongodb.py
+++ b/lib/charms/mongodb/v0/mongodb.py
@@ -145,6 +145,7 @@ class MongoDBConnection:
         try:
             for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
                 with attempt:
+                    print("ATTEMPT")
                     # The ping command is cheap and does not require auth.
                     self.client.admin.command("ping")
         except RetryError:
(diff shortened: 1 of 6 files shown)


lib/charms/mongodb/v0/mongodb.py
```diff
@@ -145,6 +145,7 @@ def is_ready(self) -> bool:
         try:
             for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
                 with attempt:
+                    print("ATTEMPT")
```
Mehdi-Bendriss: We should probably delete this

Mehdi-Bendriss (dismissed):
Thanks Mia!

Mehdi-Bendriss (approved):

PR #246, opened 2024-04-30
fix merics tests

MiaAltieri:


tests/integration/metrics_tests/test_metrics.py
```diff
@@ -38,11 +38,10 @@ async def get_address(ops_test: OpsTest, app_name=DATABASE_APP_NAME, unit_num=0)
 
 async def verify_endpoints(ops_test: OpsTest, app_name=DATABASE_APP_NAME):
     """Verifies mongodb endpoint is functional on a given unit."""
-    http = urllib3.PoolManager()
-
     for unit_id in range(len(ops_test.model.applications[app_name].units)):
```
Mehdi-Bendriss: I believe we can get away with a session object, perhaps:
```suggestion
    with requests.Session() as http:
        for unit_id in range(len(ops_test.model.applications[app_name].units)):
```

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -38,11 +38,10 @@ async def get_address(ops_test: OpsTest, app_name=DATABASE_APP_NAME, unit_num=0)
...
-        app_address = await get_address(ops_test=ops_test, app_name=app_name, unit_num=unit_id)
-        mongo_resp = http.request("GET", f"http://{app_address}:{MONGODB_EXPORTER_PORT}/metrics")
+        http = urllib3.PoolManager()
+        unit_address = await get_address(ops_test=ops_test, app_name=app_name, unit_num=unit_id)
+        mongo_resp = http.request("GET", f"http://{unit_address}:{MONGODB_EXPORTER_PORT}/metrics")
```
Mehdi-Bendriss: And we should probably add the `assert` in the loop
```suggestion
            unit_address = await get_address(ops_test=ops_test, app_name=app_name, unit_num=unit_id)
            mongo_resp = http.request("GET", f"http://{unit_address}:{MONGODB_EXPORTER_PORT}/metrics")
            assert mongo_resp.status_code == 200
```

tests/integration/metrics_tests/test_metrics.py
```diff
@@ -38,11 +38,10 @@ async def get_address(ops_test: OpsTest, app_name=DATABASE_APP_NAME, unit_num=0)
...
 
     assert mongo_resp.status == 200
```
Mehdi-Bendriss, line 49: We could perhaps delete this as it was only testing the last response

Mehdi-Bendriss (commented):
Thanks Mia! I have a couple of comments

tests/integration/metrics_tests/test_metrics.py
MiaAltieri: Actually if possible I would like to keep it as is (with my new changes from my recent commit)

my reasoning is that I want to keep tests between k8s and VM the same.

If this is a necessary change please let me know and I will make it on both charms

Mehdi-Bendriss: I would say that we should perhaps add this to the VM and K8s charm later (not urgent)

Mehdi-Bendriss (approved):
Thank you

PR #247, opened 2024-05-02
[DPE-4255] Fix tests

MiaAltieri:
# issue
1. failing metrics tests
2. failing ha tests
3. failing rel tests
4. failing backup tests
5. failing tls tests

# solution
1. fix issues with pool requests by using sessions
2. storage re-use test should run first
3. 
- relation tests do not deploy charm correctly
- checks for new user are pfrone to flakeyness 
4. it is unclear if these ever worked in the first place since they were never enabled on CI - disable these tests until they work
5. TLS tests don't allow enough time to pass before rotating certs

# future Prs
fix backup tests once scope of work is understood


tests/integration/backup_tests/test_backups.py
```diff
@@ -72,6 +71,7 @@ async def add_writes_to_db(ops_test: OpsTest):
     await clear_writes_action.wait()
 
 
+@pytest.mark.skip("Skipping tests until fixing backup tests are addressed (DPE-4264).")
```
Mehdi-Bendriss, line 74: tracking it here so we don't forget https://github.com/canonical/mongodb-k8s-operator/issues/249

Mehdi-Bendriss (approved):
Thank you!

PR #252, opened 2024-07-12
[DPE-4858] Upgrade TLS lib and add status lib

Gu1nness:
The TLS tests were artifically passing because it was impossible to integrate mongodb-k8s (multiple units) and self-signed certificates because it would get stuck in `secrets-changed` forever.

Updating the libraries, and fixing the integration between the MongoDB TLS library and the Charm's source was almost sufficient, some test methods needed to be updated as well in order to be in line with https://github.com/canonical/mongodb-operator.

A few minor improvements as well in the source charm in order to build reusable methods for pushing files to container and checking DB readiness.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,6 +27,7 @@ from charms.mongodb.v0.mongodb_backups import S3_RELATION, MongoDBBackups
 from charms.mongodb.v0.mongodb_provider import MongoDBProvider
 from charms.mongodb.v0.mongodb_secrets import SecretCache, generate_secret_label
 from charms.mongodb.v0.mongodb_tls import MongoDBTLS
+from charms.mongodb.v0.set_status import MongoDBStatusHandler
 from charms.mongodb.v0.users import (
     CHARM_USERS,
     BackupUser,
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -509,7 +532,8 @@ def _on_update_status(self, event: UpdateStatusEvent):
         # Cannot check more advanced MongoDB statuses if mongod hasn't started.
         with MongoDBConnection(self.mongodb_config, "localhost", direct=True) as direct_mongo:
             if not direct_mongo.is_ready:
-                self.unit.status = WaitingStatus("Waiting for MongoDB to start")
+                # self.unit.status = WaitingStatus("Waiting for MongoDB to start")
```
MiaAltieri: small nit
```suggestion
```

MiaAltieri (dismissed):
Amazing work Neha! looks great!

src/charm.py
Gu1nness: Indeed, I forgot to remove this line.
Done!

Mehdi-Bendriss (approved):
Thanks Neha - looks good

MiaAltieri (approved):

PR #264, opened 2024-07-16
[DPE-4874] [DPE-4868] [DPE-4870] Update libs for log rotate

MiaAltieri:
## Problem
The code for supporting log rotate is on the shared mongodb libs for `v1` - this charm uses mostly `v0` libs

## Solution
1. Update the necessary libs for log rotation to `v1`
2. Make corresponding changes in the charm such that these changes do not break the charm
3. 1+2 break tests, update them. 

## Future work
1. Update all remaining libs to their associated `v1` version
2. Implement log rotate
3. Further improve Ha tests once log rotate is fully added

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -55,3 +55,6 @@ storage:
   mongodb:
     type: filesystem
     location: /var/lib/mongodb
+  mongodb-logs:
+    type: filesystem
+    location: /var/log/mongodb
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,20 +10,11 @@ from typing import Dict, List, Optional, Set
 
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer
-from charms.mongodb.v1.helpers import (
-    build_unit_status,
-    generate_keyfile,
-    generate_password,
-    get_create_user_cmd,
-    get_mongod_args,
-)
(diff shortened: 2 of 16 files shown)


metadata.yaml
```diff
@@ -55,3 +55,6 @@ storage:
   mongodb:
     type: filesystem
     location: /var/lib/mongodb
+  mongodb-logs:
+    type: filesystem
+    location: /var/log/mongodb
```
MiaAltieri, line 60: should probably decide what we want to do here before merging

src/charm.py
```diff
@@ -1196,6 +1201,63 @@ def restart_backup_service(self) -> None:
         container = self.unit.get_container(Config.CONTAINER_NAME)
         container.restart(Config.Backup.SERVICE_NAME)
 
+    def check_relation_broken_or_scale_down(self, event: RelationDepartedEvent) -> None:
```
MiaAltieri, line 1204: this chunk was copied from MongoDB charm - maybe a future todo would be to make this into a lib?

Gu1nness (approved):

src/charm.py
Gu1nness: Agreed, but as it uses `unit_peer_data`, either we need to enforce the naming, or bring it in the library, or find another way

Mehdi-Bendriss (approved):

src/charm.py
MiaAltieri: Agreed - espcially then we can copy

PR #266, opened 2024-07-22
[DPE-4868][DPE-4869] Log rotate + tests

MiaAltieri:
## Issue
Charmed MongoDB K8s does not support log rotation

## Solution
Support log rotation using the same approach as Charmed MongoDB VM, but use pebble layer to create a reoccuring job instead of cron


templates/logrotate.j2
```diff
@@ -0,0 +1,18 @@
+{{logs_directory}}/*.log
```
MiaAltieri, line 1: copied from Charmed MongoDB VM with modifications for pebble service based of pgbouncer

tests/integration/helpers.py
```diff
@@ -508,3 +510,12 @@ async def check_or_scale_app(ops_test: OpsTest, user_app_name: str, required_uni
     count = required_units - current_units
     await ops_test.model.applications[user_app_name].scale(scale_change=count)
     await ops_test.model.wait_for_idle(apps=[user_app_name], status="active", timeout=2000)
+
+
+def audit_log_line_sanity_check(entry) -> bool:
```
MiaAltieri, line 515: copied from Charmed MongoDB VM

tests/integration/test_charm.py
```diff
@@ -106,6 +115,83 @@ async def test_application_primary(ops_test: OpsTest):
     assert number_of_primaries == 1, "more than one primary in replica set"
 
 
+@pytest.mark.group(1)
+async def test_audit_log(ops_test: OpsTest) -> None:
```
MiaAltieri, line 119: these two tests were copied from Charmed MongoDB VM

src/charm.py
```diff
@@ -218,6 +219,46 @@ def _monitor_layer(self) -> Layer:
...
+                    "startup": "enabled",
+                    "override": "replace",
+                    "backoff-delay": "1m",
+                    "backoff-factor": 1,
+                    "user": Config.UNIX_USER,
+                    "group": Config.UNIX_GROUP,
+                }
+            },
+        }
+        return Layer(layer_config)  # type: ignore
```
Gu1nness: Why is there a need of `# type: ignore` here while we return what it advertised by the type hints?

src/config.py
```diff
@@ -51,6 +53,9 @@ class LogRotate:
 
         MAX_LOG_SIZE = "50M"
         MAX_ROTATIONS_TO_KEEP = 10
+        LOG_ROTATE_TEMPLATE = "templates/logrotate.j2"
```
Gu1nness: Isn't it it safer to have a relative path here?
```suggestion
        LOG_ROTATE_TEMPLATE = "./templates/logrotate.j2"
```

Mehdi-Bendriss (dismissed):
Thank you Mia! It looks good

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #270, opened 2024-07-23
[DPE-4905] User is able to deploy Charmed MongoDB K8s as sharding components

MiaAltieri:
## Issue
Charmed MongoDB K8s can only be deployed as a replica set. It should also be able to be deployed as a `config-server` and as a `shard`

In order to support sharding using the same Charm for all sharding and non sharding components, we must allow users to configure the charm such that it starts as a shard/config-server. 

## Solution
User can deploy charm with a specified config option: “replication”, “shard”, or “config-server” and charmed MongoDB k8s will start using that role with the correct starting options.


config.yaml
```diff
@@ -5,6 +5,12 @@ options:
   auto-delete:
```
MiaAltieri, line 5: changes in this file are copied from VM charm

metadata.yaml
```diff
@@ -30,6 +30,11 @@ provides:
     interface: prometheus_scrape
```
MiaAltieri, line 30: changes in this file are copied from VM charm

src/charm.py
```diff
@@ -364,6 +395,23 @@ def users_initialized(self, value):
                 f"'users_initialized' must be a boolean value. Proivded: {value} is of type {type(value)}"
             )
 
+    @property
```
MiaAltieri, line 398: this function was copied from VM charm

src/charm.py
```diff
@@ -448,6 +498,30 @@ def is_db_service_ready(self) -> bool:
         with MongoDBConnection(self.mongodb_config, "localhost", direct=True) as direct_mongo:
             return direct_mongo.is_ready
 
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
```
MiaAltieri, line 509: this function was copied from VM charm

tests/integration/sharding_tests/test_sharding.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
MiaAltieri, line 1: changes in this file are copied from VM charm

src/charm.py
```diff
@@ -364,6 +395,23 @@ def users_initialized(self, value):
                 f"'users_initialized' must be a boolean value. Proivded: {value} is of type {type(value)}"
             )
 
+    @property
+    def role(self) -> str:
+        """Returns role of MongoDB deployment."""
+        if (
+            "role" not in self.app_peer_data
+            and self.unit.is_leader()
+            and self.model.config["role"]
```
Gu1nness, line 411: In which case would it make sense to check that `self.model.config["role"]` is not empty? Shouldn't we rather check the value against the list of supported roles (something using  `Config.Role`)?

src/charm.py
```diff
@@ -1349,3 +1423,7 @@ class SetPasswordError(Exception):
 
 if __name__ == "__main__":
     main(MongoDBCharm)
+
+
+class ShardingMigrationError(Exception):
+    """Raised when there is an attempt to change the role of a sharding component."""
```
Gu1nness: This should be higher in the code like it is in the mongodb VM charm.
The last code should be `if __name__ == "__main__":…`

Gu1nness (changes requested):

src/charm.py
MiaAltieri: This function returns the role of the charm (which can be either "replication" (default), "shard", or "config-server")

Now, FYI at any point the user can do `juju config role=new-role`, [but this functionality is not supported yet](https://github.com/canonical/mongodb-k8s-operator/blob/c3288f2a7f7523ce30fe7b7270079374e44556ad/src/charm.py#L501-L523)

This means that we **cannot trust what is in `self.model.config["role"]`** since it can change at the users command.

So what we do is:
1. trust `self.model.config["role"]` only one time which is when we set this value line 406
2. set the application data (line 406) and never reset it. Never resetting it is line 402
3. all other times return the role line 413

I hope this makes sense. I am going to embed this comment into the code that it can be more intuitive to the reader :)

Gu1nness (dismissed):

src/charm.py
```diff
@@ -448,6 +506,30 @@ def is_db_service_ready(self) -> bool:
...
             return direct_mongo.is_ready
 
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
+        """Listen to changes in application configuration.
+
+        To prevent a user from migrating a cluster, and causing the component to become
+        unresponsive therefore causing a cluster failure, error the component. This prevents it
+        from executing other hooks with a new role.
+        """
+        if self.is_role_changed():
```
Mehdi-Bendriss: nit: (to reduce one layer of nesting)
```suggestion
        if not self.is_role_changed():
            return
```

Mehdi-Bendriss: This assumes that the role has been set at deploy time. In other words, we cannot go from replication (default) to sharding

Mehdi-Bendriss (commented):
Thanks Mia! I have a question to clarify

src/charm.py
MiaAltieri: no we cannot - this isn't something we yet support 

Happy to chat about making this an epic in the future :)

Mehdi-Bendriss: Role management is coming... spec: DA116

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #271, opened 2024-07-24
[DPE-4906] - Shard components go into blocked

MiaAltieri:
## Context
In MongoDB, sharding components (config-server + shard) are not able to start unless they are “aware” of each other.

Users of juju must make these components “aware” of each other. This is done via `juju integrate` but it may not be immediately obvious to the user. 

## Solution
1. use new shared libs as necessary with this new lib:
- shards without an integrated config server go into Blocked Status: missing relation to config server
- config-servers without an integrated shard go into Blocked Status: missing relation to shard 
2. integration test that verifies this case


src/charm.py
```diff
@@ -944,6 +939,23 @@ def _get_mongodb_config_for_user(
                 tls_internal=internal_ca is not None,
             )
 
+    def _get_mongos_config_for_user(
```
MiaAltieri, line 942: copied from VM charm

tests/integration/helpers.py
```diff
@@ -48,6 +49,48 @@
 logger = logging.getLogger(__name__)
 
 
+class Status:
```
MiaAltieri, line 52: copied from VM charm

tests/integration/helpers.py
```diff
@@ -519,3 +562,88 @@ def audit_log_line_sanity_check(entry) -> bool:
             logger.error("Field '%s' not found in audit log entry \"%s\"", field, entry)
             return False
     return True
+
+
+async def get_unit_hostname(ops_test: OpsTest, unit_id: int, app: str) -> str:
```
MiaAltieri, line 565: copied from VM charm

Gu1nness (dismissed):

Gu1nness (approved):

Mehdi-Bendriss (approved):

tests/integration/helpers.py
```diff
@@ -521,6 +562,89 @@ def audit_log_line_sanity_check(entry) -> bool:
...
+        assert (
+            unit.workload_status.value == "blocked"
+        ), f"unit {unit.name} not in blocked state, in {unit.workload_status}"
+        if status:
+            assert (
+                unit.workload_status.message == status
+            ), f"unit {unit.name} not in blocked state, in {unit.workload_status}"
+
+
+async def wait_for_mongodb_units_blocked(
```
Mehdi-Bendriss, line 636: Consider making this more generic for all statuses in a future PR. Refer to [`wait_until`](https://github.com/canonical/opensearch-operator/blob/main/tests/integration/helpers_deployments.py#L284)

Mehdi-Bendriss (approved):

PR #276, opened 2024-07-26
[DPE-4904]: MongoDB raises ConnectionError when integrating TLS

Gu1nness:
* First version, but unsure this is the right solution.


Gu1nness:
Closed for now since it should be resolved by #288

PR #284, opened 2024-07-31
[DPE-4945] Password management in sharded deployments

MiaAltieri:
## Issue
Shards are able to change shared passwords (backup + operator), however since they are not the cluster brain these changes are not propagated throughout the cluster and the cluster breaks

## Solution
Do not allow shards to rotate passwords + tests to verify functionality


Mehdi-Bendriss (approved):
Thanks  Mia!

Gu1nness (approved):

PR #300, opened 2024-08-14
Update build procedure in release workflow

MiaAltieri:
## Problem
[build step in release workflow is failing](https://github.com/canonical/mongodb-k8s-operator/actions/runs/10383086858)

## Solution
Update it to match the functional build workflow from ci.yaml

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -31,8 +31,17 @@ jobs:
       contents: write  # Needed for Allure Report beta
 
   build:
-    name: Build charm
+    name: Build charm | ${{ matrix.path }}
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v19.0.0
+    strategy:
+      matrix:
+        path:
+          - .
+          - tests/integration/ha_tests/application_charm
+          - tests/integration/relation_tests/application-charm
+    with:
+      cache: true
+      path-to-charm-directory: ${{ matrix.path }}
 
   release:
       name: Release charm
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -37,8 +37,6 @@ jobs:
       matrix:
         path:
           - .
-          - tests/integration/ha_tests/application_charm
-          - tests/integration/relation_tests/application-charm
     with:
       cache: true
       path-to-charm-directory: ${{ matrix.path }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -37,9 +37,6 @@ jobs:
       matrix:
         path:
           - .
-    with:
-      cache: true
-      path-to-charm-directory: ${{ matrix.path }}
 
   release:
       name: Release charm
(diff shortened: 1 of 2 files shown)


reneradoi (dismissed):

.github/workflows/release.yaml
```diff
@@ -31,8 +31,17 @@ jobs:
...
+    name: Build charm | ${{ matrix.path }}
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v19.0.0
+    strategy:
+      matrix:
+        path:
+          - .
+          - tests/integration/ha_tests/application_charm
+          - tests/integration/relation_tests/application-charm
+    with:
+      cache: true
```
carlcsaposs-canonical: cached builds should not be used for release

carlcsaposs-canonical (dismissed):

.github/workflows/release.yaml
```diff
@@ -31,8 +31,17 @@ jobs:
...
   build:
-    name: Build charm
+    name: Build charm | ${{ matrix.path }}
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v19.0.0
+    strategy:
+      matrix:
+        path:
+          - .
+          - tests/integration/ha_tests/application_charm
+          - tests/integration/relation_tests/application-charm
```
carlcsaposs-canonical: also I don't think the test charms should be built for release

carlcsaposs-canonical:
instead, would recommend fixing build with
https://github.com/canonical/opensearch-operator/blob/8649727fcd9447fe2ee6f49de4f3ed46acd6bae0/charmcraft.yaml#L26-L28
and
https://github.com/canonical/opensearch-operator/blob/8649727fcd9447fe2ee6f49de4f3ed46acd6bae0/charmcraft.yaml#L31-L36

MiaAltieri:
@carlcsaposs-canonical thanks for the suggestion - charmcraft can be udpated in a separate PR - since I need this merged asap

reneradoi (dismissed):

carlcsaposs-canonical:
> @carlcsaposs-canonical thanks for the suggestion - charmcraft can be udpated in a separate PR - since I need this merged asap

as far as I'm aware, cached builds are not allowed for releases

also, charmcraft specifically declares that its caching is not meant to be used on releases

Gu1nness (dismissed):

reneradoi (dismissed):

Gu1nness (dismissed):

charmcraft.yaml
```diff
@@ -4,6 +4,8 @@
 type: "charm"
 parts:
   charm:
+    build-snaps:
+      - rustup
     build-packages:
```
carlcsaposs-canonical, line 9: rustc and cargo should be removed from build-packages

what is `build-essential` used for? is that mongodb-specific? if not, probably should remove that too

reneradoi (approved):

carlcsaposs-canonical (approved):

charmcraft.yaml
MiaAltieri: Unsure - from what I can tell it was added a few years ago when the charm was created , let me see if I can remove it

PR #307, opened 2024-08-20
fix: upstream-source cannot have both tag and sha

Gu1nness:
## Problem

* image is not uploaded because of `upstream-source` has both tag and sha256 which is not supported.
* Regression introduced in https://github.com/canonical/mongodb-k8s-operator/commit/4a1cbba3e07bddce5c72e57bf60bcb62f6aa0a3d

## Solution

* Remove tag as we rather use the sha256 here


MiaAltieri (approved):
thank you

delgod (approved):

Issue #301, opened 2024-08-16
Pebble ready in applications running as a Shard fail in relation "database-peers-relation-changed"


MiaAltieri:
https://github.com/canonical/mongos-k8s-operator/actions/runs/10416775957/job/28849873923?pr=20

MiaAltieri:
No longer occuring

PR #311, opened 2024-08-21
DPE-4904: Storage race condition + TLS update

Gu1nness:
## Issue

 * There's a race condition between storage ready and pebble ready event.
 * The TLS library needed to be update to ensure the db is initialised before enabling TLS

## Solution

 * Check for the storages to see if they are available
 *  Add tests about that
 * Update TLS library

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1205,6 +1205,14 @@ class MongoDBCharm(CharmBase):
         container = self.unit.get_container(Config.CONTAINER_NAME)
         container.stop(Config.SERVICE_NAME)
 
+        # Ensure the permissions are right before restarting the service.
+        try:
+            self._set_data_dir_permissions(container, all_files=True)
+
+        except (PathError, ProtocolError, MissingSecretError) as e:
+            logger.error("Cannot initialize workload: %r", e)
+            raise
+
         container.add_layer("mongod", self._mongod_layer, combine=True)
         if self.is_role(Config.Role.CONFIG_SERVER):
             container.add_layer("mongos", self._mongos_layer, combine=True)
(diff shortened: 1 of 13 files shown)


MiaAltieri (changes requested):
If this issue is a result of a race condition between TLS restart and normal start with pebble ready then I would rather fix this issue at the root of the cause which is that TLS tries to start the database first by running `restart`. Instead it should wait to run any hook by checking `db_initialised`

Gu1nness:
> If this issue is a result of a race condition between TLS restart and normal start with pebble ready then I would rather fix this issue at the root of the cause which is that TLS tries to start the database first by running `restart`. Instead it should wait to run any hook by checking `db_initialised`

Sure.
It's indeed a better solution, I'll do that and see if that works better!
I however have seen that in your case [here](https://github.com/canonical/mongodb-k8s-operator/issues/301) it happened before (in the [`mongodb-pebble-ready`](https://github.com/canonical/mongos-k8s-operator/actions/runs/10416775957/job/28849873923?pr=20#step:28:176)).
I would say the best would be to use the storage events to set the right permissions right away but I'd like your advice on it?

src/charm.py
```diff
@@ -527,6 +526,11 @@ def _on_mongod_pebble_ready(self, event) -> None:
             event.defer()
             return
 
+        if any(not storage for storage in self.model.storages.values()):
```
MiaAltieri, line 532: four things

1. excellent work digging and getting to the bottom of this
2. noob question, why do we need to check `not storage`
3. please add a comment on why this is needed so it will be obvious in the future
4. Also I think other Data Platform engineers will benefit from this knowledge! Please share if you haven't yet

tests/unit/test_charm.py
```diff
@@ -183,6 +185,22 @@ def test_pebble_ready_push_keyfile_to_workload_failure(self, push_keyfile_to_wor
             mock_container.replan.assert_not_called()
             defer.assert_called()
 
+    @patch("ops.framework.EventBase.defer")
+    def test_pebble_ready_no_storage_yet(self, defer):
```
MiaAltieri, line 189: nice

MiaAltieri (changes requested):

src/charm.py
Gu1nness: 1. Thank you :)
2. The `storage[<STORAGE-NAME>]` is not `None` when it's attached so until it's attached, this check will be True.
3. Sure
4. I will send a message to the chat yes :)

Mehdi-Bendriss (approved):
Thanks Neha! that's a good one

MiaAltieri (approved):

MiaAltieri (approved):

PR #324, opened 2024-08-28
[DPE-5289]: CI runs on new juju versions

Gu1nness:
* Use juju 3.5.3 in regular runs
 * Runs also 3.6/beta on nightly runs


carlcsaposs-canonical (dismissed):
you may need to specify base when deploying with libjuju so that it doesn't default to focal cc @shayancanonical

Gu1nness:
> you may need to specify base when deploying with libjuju so that it doesn't default to focal cc @shayancanonical

In the end I was asked to remove the nightly runs on 3.6 so it shouldn't be an issue anymore, right?

Mehdi-Bendriss (dismissed):

carlcsaposs-canonical:
> so it shouldn't be an issue anymore, right

think so!

MiaAltieri (commented):
Does test_mongos.py fail due to this bump? Or does it fail generally?

Gu1nness:
> Does test_mongos.py fail due to this bump? Or does it fail generally?

It fails generally, it is missing a retry for the `get_related_username_password`. We had the same issue on MongoDB VM a few weeks ago with `get_connection_string` if you remember.
It will be fixed when the `secret-data` ticket will be fixed.

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #327, opened 2024-09-09
[DPE-5372] Add safeguard hooks for upgrades

Gu1nness:
## Problem

* With implementation of upgrades approaching, it's good to start preparing the code.
* We need to add all safeguards for events and actions that should not be run during an upgrade.

## Solution
 * Safeguard hooks
 * Unit test for actions + events
 * (skipped) integration tests
 * Refactor some code to prevent too high complexity

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -687,9 +687,9 @@ class MongoDBCharm(CharmBase):
         if not self.db_initialised:
             return
 
-        self._reconcile_mongo_data(event)
+        self._reconcile_mongo_hosts_and_users(event)
 
-    def _reconcile_mongo_data(self, event: RelationEvent) -> None:
+    def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
         """Auxiliary function to reconcile mongo data for relation events.
 
         Args:
--- a/src/charm.py
+++ b/src/charm.py
@@ -43,6 +43,7 @@ from ops.charm import (
     CharmBase,
     ConfigChangedEvent,
     RelationDepartedEvent,
+    RelationEvent,
     StartEvent,
     UpdateStatusEvent,
 )
@@ -93,6 +94,7 @@ class MongoDBCharm(CharmBase):
         super().__init__(*args)
 
         self.framework.observe(self.on.mongod_pebble_ready, self._on_mongod_pebble_ready)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.update_status, self._on_update_status)
         self.framework.observe(
@@ -654,16 +656,23 @@ class MongoDBCharm(CharmBase):
             event.defer()
             return
 
-    def _relation_changes_handler(self, event) -> None:
+    def _relation_changes_handler(self, event: RelationEvent) -> None:
         """Handles different relation events and updates MongoDB replica set."""
         self._connect_mongodb_exporter()
         self._connect_pbm_agent()
 
-        if type(event) is RelationDepartedEvent:
+        if isinstance(event, RelationDepartedEvent):
             if event.departing_unit.name == self.unit.name:
                 self.unit_peer_data.setdefault("unit_departed", "True")
 
         if not self.unit.is_leader():
+            return
+
+        if self.upgrade_in_progress:
+            logger.warning(
+                "Adding replicas during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
+            )
+            event.defer()
             return
 
         # Admin password and keyFile should be created before running MongoDB.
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -659,25 +661,40 @@ def _on_start(self, event) -> None:
...
         # Admin password and keyFile should be created before running MongoDB.
         # This code runs on leader_elected event before mongod_pebble_ready
         self._generate_secrets()
 
         if not self.db_initialised:
             return
 
+        self._reconcile_mongo_data(event)
+
+    def _reconcile_mongo_data(self, event: RelationEvent) -> None:
```
MiaAltieri:
```suggestion
        self._reconcile_mongo_hosts(event)

    def _reconcile_mongo_hosts(self, event: RelationEvent) -> None:
```

src/charm.py
```diff
@@ -93,6 +94,7 @@ def __init__(self, *args):
         super().__init__(*args)
 
         self.framework.observe(self.on.mongod_pebble_ready, self._on_mongod_pebble_ready)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
```
MiaAltieri, line 97: 🤦‍♀️ ty

tests/integration/upgrades/test_upgrades.py
```diff
@@ -0,0 +1,61 @@
...
+    current_password = await get_password(ops_test, leader_id, app_name=app_name)
+    new_charm = await ops_test.build_charm(".")
+    await ops_test.model.applications[app_name].refresh(path=new_charm)
+    results = await set_password(ops_test, leader_id, password="0xdeadbeef", app_name=app_name)
+
+    assert results == "Cannot set passwords while an upgrade is in progress."
+
+    after_action_password = await get_password(ops_test, leader_id, app_name=app_name)
+
+    assert current_password == after_action_password
```
MiaAltieri, line 58: 🤙 sick I like this initiative ty

MiaAltieri (approved):

tests/integration/upgrades/test_upgrades.py
```diff
@@ -0,0 +1,61 @@
...
+    await ops_test.model.deploy(
+        new_charm,
+        resources=resources,
+        application_name=app_name,
+        num_units=3,
+        series="jammy",
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[app_name], status="active", timeout=1000, idle_period=120
+    )
```
MiaAltieri: please deploy charm from charmhub here

MiaAltieri (changes requested):

src/charm.py
Gu1nness: Thing is, it does not only reconcile mongo hosts but also users (as part of `update_app_relation_data`)

MiaAltieri: hmm then maybe `_reconcile_mongo_hosts_and_users`? I think `data` is confusing here

`metadata` could also be good but I also think word is also vague

Gu1nness: Renamed to `_reconcile_mongo_hosts_and_users`

Mehdi-Bendriss (approved):

tests/integration/upgrades/test_upgrades.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_upgrade_password_change_fail(ops_test: OpsTest):
+    app_name = await get_app_name(ops_test)
+    leader_id = await find_unit(ops_test, leader=True, app_name=app_name)
+
+    current_password = await get_password(ops_test, leader_id, app_name=app_name)
+    new_charm = await ops_test.build_charm(".")
+    await ops_test.model.applications[app_name].refresh(path=new_charm)
+    results = await set_password(ops_test, leader_id, password="0xdeadbeef", app_name=app_name)
```
MiaAltieri, line 52: just FYI I am not sure if we will be able to guarantee that an upgrade is running *while* running the action `set_password`

the reason is, is that if the snap revision doesn't change then the UX is simply `juju refresh` and the refresh happens instantaneously (or very quickly for that matter) 

if the snap revision *does* change then the UX is `juju refresh; juju resume-upgrade` and in between those two actions we can guarantee that an upgrade is running and we could do  `juju refresh; juju run mongodb-k8s/leader set-password; juju resume-upgrade`

So I am not sure actually if it makes sense to have this here or not. But I would love to hear your thoughts and expertise

Gu1nness: That's exactly what I was wondering.
I added this because it was part of the DoD of [the ticket](https://warthogs.atlassian.net/browse/DPE-5372).
If we go to upgrade V3, it will be easier to do that because we can set `pause-after-refresh` to `all`.
I did this (fully skipped) as a first version to be updated when we would have upgrades V3.

MiaAltieri (approved):

PR #328, opened 2024-09-09
[DPE-5215] Upgrade libraries to fix secret disclosure issue

Gu1nness:
## Problem

* We were using the `DatabaseProvider` class wrong
* This lead to a secret disclosure issue with mongo-k8s

## Solution

* The library was fixed in https://github.com/canonical/mongodb-operator/pull/472
* Bump here + small fixes

## Miscellaneous

* Fix `db_initialised` property

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -11,10 +11,11 @@ from typing import Optional
 
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseProvides,
+    DatabaseRequestedEvent,
     DatabaseRequires,
 )
 from charms.mongodb.v1.mongos import MongosConnection
-from ops.charm import CharmBase, EventBase, RelationBrokenEvent
+from ops.charm import CharmBase, EventBase, RelationBrokenEvent, RelationChangedEvent
 from ops.framework import Object
 from ops.model import (
     ActiveStatus,
@@ -42,7 +43,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 12
 
 
 class ClusterProvider(Object):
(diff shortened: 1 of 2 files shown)


lib/charms/mongodb/v1/helpers.py
```diff
@@ -320,3 +320,25 @@ def add_args_to_env(var: str, args: str):
...
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            shell=isinstance(command, str),
+            env=env,
+            cwd=working_dir,
+        )
+        logger.debug(f"{output=}")
+        return output
+    except subprocess.CalledProcessError as err:
+        logger.error(f"cmd failed - {err.cmd = }, {err.stdout = }, {err.stderr = }")
```
MiaAltieri, line 343: does this formatting pass the linter? it was causing me issues on one of the charms

MiaAltieri (dismissed):

MiaAltieri (dismissed):

src/charm.py
```diff
@@ -339,7 +344,7 @@ def app_peer_data(self) -> RelationDataContent:
     @property
     def db_initialised(self) -> bool:
         """Check if MongoDB is initialised."""
-        return "db_initialised" in self.app_peer_data
+        return json.loads(self.app_peer_data.get("db_initialised", "false").lower())
```
MiaAltieri: IIRC .lower was needed to be compatible with older versions is that correct?

if so you can remove it here. Since we do not support upgrades on K8s we don't have to worry about this sort of compatibility

MiaAltieri: actually I got it handled in [this PR](https://github.com/canonical/mongodb-k8s-operator/pull/329/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307c)

Gu1nness: Yeah it should be okay I'll give it a try

Gu1nness: Handled in your PR, removing it from here :)

MiaAltieri (dismissed):

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #330, opened 2024-09-12
tests: remove unnecessary patching of private ops class

tonyandrewmeyer:
The tests currently patch `ops.testing._TestingModelBackend`, which is a private class. This will break in an upcoming release of ops.

The patching is actually not required any more, so can just be entirely removed rather than updated.

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -23,8 +23,6 @@ from tenacity import stop_after_attempt, wait_fixed, wait_none
 
 from charm import MongoDBCharm, NotReadyError
 
-from .helpers import patch_network_get
-
 PYMONGO_EXCEPTIONS = [
     (ConnectionFailure("error message"), ConnectionFailure),
     (ConfigurationError("error message"), ConfigurationError),
@@ -36,7 +34,6 @@ logger = logging.getLogger(__name__)
 
 
 class TestCharm(unittest.TestCase):
-    @patch_network_get(private_address="1.1.1.1")
     def setUp(self):
         self.maxDiff = None
         self.harness = Harness(MongoDBCharm)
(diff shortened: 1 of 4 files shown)


MiaAltieri (dismissed):
Thank you so much for this change @tonyandrewmeyer - Really appreciate your foresight here, I notified the DP team + I've made GitHub issues in the other Mongo* repos as to not forget about it


- [MongoDB VM](https://github.com/canonical/mongodb-operator/issues/485)
- [Mongos VM](https://github.com/canonical/mongos-operator/issues/56)
- [Mongos K8s](https://github.com/canonical/mongos-k8s-operator/issues/36)

tests/unit/test_charm.py
```diff
@@ -1044,7 +1041,6 @@ def test_set_password_provided(self, connection):
         # verify app data is updated and results are reported to user
         self.assertEqual("canonical123", new_password)
 
-    @patch_network_get(private_address="1.1.1.1")
```
phvalguima, line 1050: Some of our charms are using this specific value `1.1.1.1` to do further checks, e.g. [here](https://github.com/canonical/opensearch-operator/blob/f8b701f2980d5d56e6b93541ec1c39d9e65d602b/tests/unit/lib/test_opensearch_peer_clusters.py#L131).

How can we now mock the `network_get` and return a specific private_address?

phvalguima (commented):
Hi @tonyandrewmeyer really appreciate your attention here. Thanks for highlighting the new changes in `ops` and how it will affect the team and the charms!

Left just a question, as some charms' tests do depend on a specific return from a mocked `network_get`. Can you just explain how should it work from now on?

tonyandrewmeyer:
> Left just a question, as some charms' tests do depend on a specific return from a mocked `network_get`. Can you just explain how should it work from now on?

When adding a relation, if there's no global network configured, then Harness checks for one specific to the (relation name, relation ID) and will add 10.0.0.10 if one doesn't exist, and checks for one specific to just the relation name and will add 192.0.2.0 if one doesn't exist.

If you need different addresses, then you can just call [Harness.add_network](https://ops.readthedocs.io/en/latest/#ops.testing.Harness.add_network), making sure you do this *before* adding the relation.

Calling with just the address (like `harness.add_network('192.0.2.1')`) sets up the default binding, or you can call with the `endpoint` (relation name) and `relation_id` set if it should be specific to those.

PR #341, opened 2024-10-08
[DPE-5649][DPE-5647][DPE-5645] update mongo* libs

MiaAltieri:
## Issue
1. mongo* libs need to be updated with latest fixes
2. tests are failing due to
- ops deprecation notice getting shown in `juju run-action` results :/
- statuses are updated
- mongos client is not initialised properly 

## Solution
1. charmcraft-fetch
2. update tests for associated issues for a green CI

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, List, Optional
 
 import yaml
 from dateutil.parser import parse
+from more_itertools import one
 from pytest_operator.plugin import OpsTest
 from tenacity import Retrying, retry, stop_after_attempt, stop_after_delay, wait_fixed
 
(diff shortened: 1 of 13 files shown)


tests/integration/helpers.py
```diff
@@ -528,7 +528,7 @@ def get_password_using_subprocess(
         logger.error("get-password command returned non 0 exit code: %s", result.stderr)
         raise Exception(f"get-password command returned non 0 exit code: {result.stderr}")
     try:
-        password = result.stdout.decode("utf-8").split("password:")[-1].strip()
```
Gu1nness, line 531: I have a better version of that [here](https://github.com/canonical/mongodb-k8s-operator/blob/DPE-5374-mongo-db-k-8-s-upgrades-test-existing-vm-code-of-pre-upgrades-hook-on-k-8-s/tests/integration/helpers.py#L509-L549) which uses the json format to ensure that is will work whatever happens :)

Gu1nness (approved):

phvalguima (approved):
LGTM

I am missing the context on the lib update, but I believe this is coming from the VM charm instead of being changes in this charm.

PR #348, opened 2024-10-14
DRAFT + WIP [DPE-5661] patch stateful set to ensure pod isn't removed prematurely 

MiaAltieri:
# Issue
1. juju only gives us 30 seconds to clean up a unit being removed

# Solution
1. patch stateful set to have a very long termination period


src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
+                    # remove_replset_member retries for 60 seconds
+                    with MongoDBConnection(self.mongodb_config) as mongo:
+                        mongo.remove_replset_member(self.unit_host(self.unit))
+
+        except NotReadyError:
+            logger.info(
+                "Failed to remove %s from replica set, another member is syncing", self.unit.name
+            )
+        except PyMongoError as e:
+            logger.error("Failed to remove %s from replica set, error=%r", self.unit.name, e)
```
MiaAltieri, line 867: Solves [DPE-5665] - completely separate from patching stateful set. Instead it properly handles scale down

Copy and paste from MongoDB VM

src/charm.py
```diff
@@ -42,6 +42,9 @@
     CrossAppVersionChecker,
     get_charm_revision,
 )
+from lightkube import Client
+from lightkube.resources.apps_v1 import StatefulSet
+from lightkube.types import PatchType
```
Gu1nness, line 53: Need to import lightkube in pyproject.toml

src/charm.py
```diff
@@ -646,6 +660,19 @@ def _on_start(self, event) -> None:
...
+        # https://bugs.launchpad.net/juju/+bug/2035102
+
+        # It doesn't matter if we patch the stateful set before or after the charm has started.
+        # The usual start hooks emitted by juju will have already been emitted, so we can expect
+        # two rounds of restarts on one or more units (some units that get initialised late will
+        # only have one round of restarts). The second round of start hooks will be emitted
+        # **only after the replica set has been initialized**, we have 0 control over that.
+
+        if self.unit.is_leader() and self.get_current_termination_period() != ONE_YEAR:
+            self.update_termination_grace_period(ONE_YEAR)
```
Gu1nness: we already have an elected leader here right?

src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
-                time.sleep(1)
-                timeout -= 1
-                if timeout < 0:
-                    raise Exception(f"{self.unit.name}.on_stop timeout exceeded")
-            logger.debug(f"{self.unit.name} releasing on_stop")
-            self.unit_peer_data["unit_departed"] = ""
+    def get_current_termination_period(self) -> int:
+        """Returns the current termination period for the stateful set of this juju application."""
+        client = Client()
+        statefulset = client.get(StatefulSet, name=self.app.name, namespace=self.model.name)
```
Gu1nness, line 893: Not sure we need to specify the namespace here but should not be an issue to have it.

src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
+        # can result in data-loss
+        if not self.is_role(Config.Role.SHARD):
+            return
+
+        client = Client()
+        patch_data = {
+            "spec": {
+                "template": {
+                    "spec": {"terminationGracePeriodSeconds": ONE_YEAR},
+                    "metadata": {"annotations": {"force-update": str(int(time.time()))}},
```
Gu1nness, line 908: What it the force-update annotation useful for here?

src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
+        # updating the termination grace period is only useful for shards, whose sudden removal
+        # can result in data-loss
+        if not self.is_role(Config.Role.SHARD):
+            return
+
+        client = Client()
+        patch_data = {
+            "spec": {
+                "template": {
+                    "spec": {"terminationGracePeriodSeconds": ONE_YEAR},
```
Gu1nness, line 907: Isn't one year a bit too much?
It means that your pod can take up to 1 year to be removed.
I would keep it lower, up to a day?

src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
+        client = Client()
+        statefulset = client.get(StatefulSet, name=self.app.name, namespace=self.model.name)
+        return statefulset.spec.template.spec.terminationGracePeriodSeconds
+
+    def update_termination_grace_period(self, seconds: int) -> None:
+        """Patch the termination grace period for the stateful set of this juju application."""
+        # updating the termination grace period is only useful for shards, whose sudden removal
+        # can result in data-loss
+        if not self.is_role(Config.Role.SHARD):
+            return
```
MiaAltieri, line 901: im unsure about this bit

src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
+                "template": {
+                    "spec": {"terminationGracePeriodSeconds": ONE_YEAR},
+                    "metadata": {"annotations": {"force-update": str(int(time.time()))}},
+                }
+            }
+        }
+        client.patch(
+            StatefulSet,
+            name=self.app.name,
+            namespace=self.model.name,
```
Gu1nness, line 915: Same here, not sure the namespace is needed but it should cause an issue to have it either

src/charm.py
```diff
@@ -759,19 +782,107 @@ def _reconcile_mongo_hosts_and_users(self, event: RelationEvent) -> None:
...
+            obj=patch_data,
+            patch_type=PatchType.MERGE,
```
Gu1nness, line 917: I think MERGE or STRATEGIC don't change anything here so I would not specify it and keep the default value of `STRATEGIC`, but it's a nitpick.

Gu1nness (dismissed):
Overall it should be working.
I would try to leave it to run and play with it to see how stable it stays in the time

Gu1nness (commented):
Overall it should be working.
I would try to leave it to run and play with it to see how stable it stays in the time

src/charm.py
MiaAltieri: yes we do! <3

src/charm.py
MiaAltieri: I chose a ridiculously long amount on purpose to ensure that shard drainage would have enough time in worst case scenarios. 

i.e. if there are jumbo-chunks that we cannot automatically drain (because the user should decide the new shard key) - then that will require user intervention and I would like to give them more than enough time to do so.

PR #349, opened 2024-10-15
fix: bump mongodb_backups

Gu1nness:
## Issue

The dictionary can have a None value in case there's no snapshot: pbm status can contain a null value for pbm[backups][snapshot] which is parsed as None and then pbm_status["backups"].get("snapshot", []) == None which is not what is expected here since we want to iterate on the results.

## Solution

Export the fallback out of the get in order to ensure that we can iterate on it later.


Mehdi-Bendriss (approved):
Could you please add a short description for future readers?

MiaAltieri (approved):

PR #352, opened 2024-10-17
feat: Mutating webhook

Gu1nness:
## Issue

With big databases we need to have a long time to shutdown our databases.
Juju only leaves 30s to do that which is way too low.

With this PR, we deploy a sidecar container that serves mutating and validation webhook endpoints.
We then create the service for this webhook and the mutating endpoint associated with this service.
We take care of removing it at the end and to keep the certificates so all (fastapi) services can answer on that endpoint.

### DIFF
--- a/src/gen_cert.py
+++ b/src/gen_cert.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+"""Generates a self signed certificate for the mutating webhook."""
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+import datetime
+
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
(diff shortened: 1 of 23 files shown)


src/gen_cert.py
```diff
@@ -0,0 +1,52 @@
...
+    private_key = rsa.generate_private_key(
+        public_exponent=65537, key_size=2048, backend=default_backend()
+    )
+    public_key = private_key.public_key()
+
+    builder = x509.CertificateBuilder()
+    builder = builder.subject_name(x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, app_name)]))
+    builder = builder.issuer_name(x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, ns)]))
+    builder = builder.not_valid_before(datetime.datetime.today() - one_day)
+    builder = builder.not_valid_after(datetime.datetime.today() + (one_day * 365 * 100))
```
MiaAltieri, line 26: nice :)

PR #373, opened 2025-01-06
[DPE-6191] fix install timeout + fix broken CI

MiaAltieri:
# Issue

No uniform timeout for deployment installation in tests

Multiple CI workflows are failing

# Solution

As in VM charms, add a uniform constant for deployment timeout

to fix build
- update poetry
- address new lint issues
- use new DP pipelines


pyproject.toml
```diff
@@ -2,16 +2,11 @@
...
-name = "mongodb-k8s"
-version = "0.0.1-dev.0"
-description = ""
-license = "Apache-2.0"
-authors = [
-    "Mykola Marzhan <mykola.marzhan@canonical.com>"
-]
-readme = "README.md"
-homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
-repository = "https://github.com/canonical/mongodb-k8s-operator"
```
Gu1nness, line 14: I think we need to move this information to the new [project](https://python-poetry.org/blog/announcing-poetry-2.0.0#supporting-the-project-section-in-pyprojecttoml-pep-621) section so we don't lose this information :)

(See the other suggestion comment)

pyproject.toml
```diff
@@ -2,16 +2,11 @@
...
+package-mode = false
+requires-poetry = ">=2.0.0"
+
+
+
```
Gu1nness, line 9: I suggest this so we don't lose the metadata about the project.

```suggestion
package-mode = false
requires-poetry = ">=2.0.0"


[project]
name = "mongodb-k8s"
version = "0.0.1-dev.0"
description = ""
authors = [
    {name = "Mykola Marzhan", email = "mykola.marzhan@canonical.com"}
]
readme = "README.md"
requires-python = ">=3.10"

[project.urls]
homepage = "https://charmhub.io/mongodb-k8s?channel=6/edge"
repository = "https://github.com/canonical/mongodb-k8s-operator"
```

src/config.py
```diff
@@ -22,7 +22,7 @@ class Config:
     MONGOD_CONF_DIR = "/etc/mongod"
     MONGODB_LOG_FILENAME = "mongodb.log"
 
-    CHARM_INTERNAL_VERSION_FILE = "charm_internal_version"
+    CHARM_INTERNAL_VERSION_FILE = "charm_version"
```
Gu1nness, line 25: :purple_heart: 

Hopefully we soon can publish a new version of Data Helpers which removes the difference between charm internal version and charm version since it's not meaningful anymore.

tests/integration/upgrades/test_local_sharding_upgrades.py
```diff
@@ -186,7 +187,7 @@ async def test_upgrade_cluster(ops_test: OpsTest, righty_upgrade_charm, add_writ
     assert (
         actual_shard_two_writes == shard_two_total_expected_writes
     ), "missed writes during upgrade procedure."
-    logger.error(f"{actual_shard_one_writes = }, {actual_shard_two_writes = }")
+    logger.error(f"{actual_shard_one_writes=}, {actual_shard_two_writes=}")
```
Gu1nness: Nit: I liked it better with the spaces around which makes it easier to read

Gu1nness (changes requested):

Gu1nness:
Final comment, do you think we also can take care of the moving to charmcraft 3 that Carl proposed or not?

tests/integration/upgrades/test_local_sharding_upgrades.py
MiaAltieri: @Gu1nness I agree but it causes a linting error , let me add a ` # noqa: E226`

MiaAltieri:
@Gu1nness : 

> Final comment, do you think we also can take care of the moving to charmcraft 3 that Carl proposed or not?

I believe that Carl can do this as part of [this comment thread](https://chat.canonical.com/canonical/pl/wekd7r4eqtdn7mxhgxgej8zzsr) - we just need to be sure that we meet all qualifications

pyproject.toml
MiaAltieri: This pyproject.yaml was created before poetry became the standard of Data Platform charms, now that poetry is the standard, these fields are no longer used. I'd rather keep with the scheme from the other repos, but if you feel strongly just let me know and I can add this in!

Gu1nness: Oh no it's okay if it's the standard elsewhere, go for it and keep it as is

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

PR #374, opened 2025-01-09
[DPE-6303] add simple replica set terraform

MiaAltieri:
## Issue
We would like an easy way to deploy MongoDB using terraform on k8s

## Solution
Add terraform using the same approach as MongoDB VM

## Manual Testing
```
terraform init
terraform plan -out myplan
terraform apply "myplan"
```
done on microk8s


.github/workflows/ci.yaml
Gu1nness, line 1: If this is something we start doing often, we might want to include it in DPW in the future

terraform/tests/variables.tf
```diff
@@ -0,0 +1,16 @@
+variable "model_name" {
+  description = "Model name"
+  type        = string
+}
+
+variable "app_name" {
+  description = "mongodb app name"
```
Gu1nness:
```suggestion
  description = "MongoDB app name"
```

Gu1nness (dismissed):

Gu1nness (approved):

terraform/main.tf
```diff
@@ -0,0 +1,41 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+resource "juju_application" "mongodb-k8s" {
+
+  charm {
+    name     = "mongodb-k8s"
+    channel  = var.channel
+    revision = var.revision
+    base     = "ubuntu@22.04"
```
Mehdi-Bendriss, line 10: in the future, this will be extracted in its own variable when we support dual bases

Mehdi-Bendriss (approved):
Thanks Mia!

PR #380, opened 2025-01-20
[DPE-6347] full replica set tf

MiaAltieri:
# Issue
Missing all possible integrations for replica sets

# Solution
Add in all remaining integrations

# Other
Fo mongodb VM terraform we integrate with `grafana-agent` as it is a requirement of the spec. However, VM charms do not integrate with `grafana-agent`. Instead they expose COS endpoints for loki, prometheus, and grafana-dashboards. The spec specifies that complete solutions for terraform bundles should expose these endpoints, so there is nothing for us to add here


Mehdi-Bendriss (dismissed):

terraform/modules/variables.tf
```diff
@@ -4,7 +4,7 @@ variable "model_name" {
 }
 
 variable "app_name" {
-  description = "MongoDB app name"
+  description = "mongodb app name"
```
Gu1nness: Why has this been lowercased?

MiaAltieri: good catch

.github/workflows/ci.yaml
Gu1nness, line 1: You need to update the terraform path l.105 in the terraform deploy

MiaAltieri: good catch ty

.github/workflows/ci.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
           juju model-defaults logging-config='<root>=INFO; unit=DEBUG'
           juju add-model test
       - name: Terraform deploy
-        working-directory: ./terraform/tests/
+        working-directory: ./terraform/modules/
```
Gu1nness: There's an inconsistency in this working directory path and the former one :)
```suggestion
        working-directory: ./terraform/modules
```

Gu1nness (dismissed):

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #381, opened 2025-01-20
[DPE-6349] add sharded tf

MiaAltieri:
## Issue
SQA would like a repeatable way to deploy our sharded mongodb

## Solution
Deploy a complete sharded cluster with all available integrations.

## Other
Fo mongodb VM terraform we integrate with `grafana-agent` as it is a requirement of the spec. However, VM charms do not integrate with `grafana-agent`. Instead they expose COS endpoints for loki, prometheus, and grafana-dashboards. The spec specifies that complete solutions for terraform bundles should expose these endpoints, so there is nothing for us to add here


.github/workflows/ci.yaml
```diff
@@ -80,7 +80,31 @@ jobs:
           terraform init
           terraform fmt
           terraform validate
-      - name: run checks - prepare
+
```
MiaAltieri, line 83: same workflow as MongoDB vm

terraform/modules/sharded_cluster/preamble.tf
```diff
@@ -0,0 +1,43 @@
+resource "null_resource" "preamble" {
```
MiaAltieri, line 1: these files in the directory are almost identical to mongodb vm version

terraform/modules/sharded_cluster/sharded_deployment.tf
```diff
@@ -0,0 +1,141 @@
...
+
+}
+
+resource "juju_integration" "tls-operator_mongodb-integration" {
+  for_each = merge(
+    local.mongodb_apps,
+    {
+      "mongos-k8s" = {
+        app_name = "mongos-k8s"
+        units    = 1
```
Mehdi-Bendriss, line 95: Can you explain the units count here?

terraform/modules/sharded_cluster/variables.tf
```diff
@@ -0,0 +1,31 @@
...
+variable "config_server_app_name" {
+  description = "config-server app name"
+  type        = string
+  default     = "config-server"
+}
+
+variable "config_server_replicas" {
+  description = "Node count"
+  type        = number
+  default     = 1
```
Mehdi-Bendriss: We should perhaps for this one and for the subsequent shards' replicas set the default to 3 as recommended number

Mehdi-Bendriss (commented):
Thanks Mia! Solid work! I left 2 questions

terraform/modules/sharded_cluster/sharded_deployment.tf
MiaAltieri: Good question - actually no. Just habit I suppose, in our integration tests and the tutorial we deploy one unit of mongos. A user can use more than one mongos router when they have a load that demands it

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #385, opened 2025-01-30
[DPE-6349] fix sharded tf

MiaAltieri:
# Issue
Terraform was incorrect, since it used the name of the application for the role.

# Solution
Use the correct field for role


paulomach (approved):
outstanding

reneradoi (approved):
Agree with @paulomach

PR #401, opened 2025-04-25
feat(DPE-6723): LDAP Support

Gu1nness:
This PR bumps the mongo single kernel library to support LDAP.


tests/integration/sharding_tests/helpers.py
```diff
@@ -62,44 +64,105 @@ async def deploy_cluster_components(
...
-    )
-
-    await ops_test.model.wait_for_idle(
-        apps=CLUSTER_COMPONENTS,
-        idle_period=20,
-        timeout=TIMEOUT,
-        raise_on_blocked=False,
-        raise_on_error=False,
-    )
+    if channel is None:
```
Mehdi-Bendriss, line 67: why is the code so duplicated?

tests/unit/test_charm.py
```diff
@@ -116,34 +116,8 @@ def test_mongod_pebble_ready(
...
-                            "--logRotate reopen --logappend --logpath=/var/log/mongodb/mongodb.log "
-                            "--auditDestination=file "
-                            "--auditFormat=JSON "
-                            "--auditPath=/var/log/mongodb/audit.log "
-                            "--replSet=mongodb-k8s "
-                            f"--dbpath={self.harness.charm.workload.paths.data_path}"
-                        )
-                    },
+                    "command": "/bin/bash /bin/start-mongod.sh",
+                    "environment": {"MONGOD_ARGS": ""},
```
Mehdi-Bendriss, line 120: are mongod args empty when starting mongodb due to the file config?

Mehdi-Bendriss (commented):
Thanks Neha, the if/else channel change is not clear to me

tests/unit/test_charm.py
Gu1nness: Yes !

tests/integration/sharding_tests/helpers.py
Gu1nness: It's not duplicated, but there's a bug currently (fixed by this release) that makes the deployment fail to init replicaset from time to time so in order to prevent that we deploy with one and then scale up.
This will be reverted right after that release with the new code that fixes this bug.

Mehdi-Bendriss (approved):

PR #410, opened 2025-07-17
feat(DPE-7739): Remove integration tests from charm

Gu1nness:
## Issue

Prepare new stable release.

## Solution

Remove integration tests as they where moved to single kernel library


Mehdi-Bendriss (approved):

tests/integration/test_charm.py
```diff
@@ -85,475 +58,3 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
     await ops_test.model.set_config({"update-status-hook-interval": "60m"})
-
-
-@pytest.mark.group(1)
-@pytest.mark.abort_on_fail
-@pytest.mark.parametrize("unit_id", UNIT_IDS)
-async def test_application_is_up(ops_test: OpsTest, unit_id: int):
-    address = await get_address_of_unit(ops_test, unit_id=unit_id)
-    response = MongoClient(address, directConnection=True).admin.command("ping")
-    assert response["ok"] == 1
```
Mehdi-Bendriss: I'd rather this stays (like we have for mongos)
The same applies to sharding - basically ensuring mongodb is up in addition to the juju units

Mehdi-Bendriss (changes requested):

tests/integration/test_charm.py
Gu1nness: Ok !

Mehdi-Bendriss (approved):

PR #423, opened 2025-08-29
patch: remove auto-delete config option

patriciareinoso:
## Issue
The auto-delete config option is no longer a desired config option in release 8

## Solution
This PR removes the auto-delete config option and its logic.
Now, the databases are never dropped.
This prevents unintended data loss in case of misconfiguration.


Mehdi-Bendriss (dismissed):

Gu1nness:
You also need to bump the library version to the new version containing your patch in your PR to remove the workflows for that option.

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #427, opened 2025-09-26
feat: Terraform Modules

Gu1nness:
* Terraform module for replica set
* Terraform module for sharded cluster
* Terraform module for product replicaset
* Terraform module for product sharded cluster

Change wrt VM module:
 * Remove expose config (makes a bit harder to test)
 * Remove condition of data integrator being on the same model as mongos k8s
 * rename all mongodb / mongos to mongodb-k8s / mongos-k8s


terraform/charm/replica_set/variables.tf
```diff
@@ -1,5 +1,7 @@
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
```
patriciareinoso:
```suggestion
```

terraform/charm/replica_set/variables.tf
```diff
@@ -46,18 +52,23 @@ variable "constraints" {
...
 
 variable "storage" {
   description = "Map of storage used by the application"
   type        = map(string)
   default     = {}
+
+  validation {
+    condition     = length(var.storage) == 0 || lookup(var.storage, "count", 0) <= 1
+    error_message = "Only one storage is supported"
+  }
```
patriciareinoso, line 65: don't we have separation of storages in mongodb-k8s already?

terraform/product/sharded/integrations.tf
```diff
@@ -0,0 +1,124 @@
+# Copyright 2024 Canonical Ltd.
```
patriciareinoso, line 1: isn't the `logging` integration missing here? (same for the replica set)

patriciareinoso (commented):
LGTM. 
I just added a couple of questions

terraform/charm/replica_set/variables.tf
Gu1nness: Indeed, fixing that.

terraform/product/sharded/integrations.tf
Gu1nness: I'm not sure, those integrations are not handler by the deployment, but exposed as part of the outputs to integrate with COS terraform module.

Mehdi-Bendriss (approved):
Thanks Neha!

patriciareinoso (approved):

PR #537, opened 2026-04-10
fix: add dockerhub mirror

Gu1nness:


reneradoi (approved):

spread.yaml
```diff
@@ -115,6 +117,30 @@ prepare: |
   cd "$SPREAD_PATH"
   snap install --classic concierge
 
+  if [[ -n "$DOCKERHUB_MIRROR" ]]
+  then
+    # Running on IS-hosted runner; configure microk8s to use Docker Hub mirror
+    # Run before concierge prepare because of https://github.com/canonical/concierge/issues/75
+
+    snap install microk8s --channel "$CONCIERGE_MICROK8S_CHANNEL" --classic
```
carlcsaposs-canonical, line 125: fyi it looks like concierge supports mirrors natively now:
https://github.com/canonical/concierge?tab=readme-ov-file#schema
https://github.com/canonical/concierge/issues/75

carlcsaposs-canonical (approved):

spread.yaml
Gu1nness: I'll test that a bit later but that looks promising!

